PR #8, opened 2021-10-18
Start tests

ianlongden:
Trying to do this in small increments. So it is easier to review. So no way near complete but a basic idea of what is intended should be seen.

### DIFF
--- a/backend/app/literature/config.py
+++ b/backend/app/literature/config.py
@@ -71,13 +71,10 @@ class FactoryConfig:
 
     def __call__(self):
         if self.env_state == "build":
-            print("Calling DEV")
             return DevConfig()
         elif self.env_state == "prod":
-            print("Calling Prod")
             return ProdConfig()
         elif self.env_state == "test":
-            print("Calling Test")
             return TestConfig()
 
 


backend/app/main.py
```diff
@@ -3,18 +3,18 @@
 import argparse
 
 from uvicorn.config import LOGGING_CONFIG
-from starlette.graphql import GraphQLApp
+# from starlette.graphql import GraphQLApp
```
adamjohnwright, line 6: These commented out lines can be deleted if you are sure we don't need them. This one we for sure don't need as it was from when I was playing with getting GraphQL working but realized it was too much to take on at the beginning at least.

backend/app/literature/schemas/resource_schemas.py
```diff
@@ -2,7 +2,7 @@
 from datetime import datetime
 
 from pydantic import BaseModel
-from pydantic import ValidationError
+# from pydantic import ValidationError
```
adamjohnwright, line 5: Can delete this is not used

backend/app/literature/schemas/reference_schemas.py
```diff
@@ -52,6 +52,18 @@ class Config():
         orm_mode = True
         extra = "forbid"
 
+    @validator('title')
```
adamjohnwright, line 55: This is needing to be cleaned up a little. The fact that the title and category are not listed as optional means that it will never be None. I guess what you are adding is that the title also can't be a blank string. The category doesn't need the validator because it has to be populated with one of the options from the enum. In general this type of validation is only required for the PATCH enpoints... as the required fields in the POST should not be None if they are supplied in the PATCH. Hopefully that makes sense.

README.md
```diff
@@ -503,3 +503,17 @@ aws_secret_access_key = `
 - Proceed with the appropriate make commands as usual.
 - Reminder: this process needs to be repeated every time you get an error like this (usually ~ every 12 hours):
 `Error response from daemon: pull access denied for 100225593120.dkr.ecr.us-east-1.amazonaws.com/agr_neo4j_env, repository does not exist or may require 'docker login': denied: Your authorization token has expired. Reauthenticate and try again.`
+
+
+## testing (Tempory, will removed once it is all sorted properly)
+
+# kick of the postgres db
+docker-compose -f docker-compose-test.yml up
```
adamjohnwright, line 511: I would surround all these command line commands with 

```bash 
<your command(s)
```

azurebrd:
Cool cool

PR #9, opened 2021-10-18
tests for resource and missing file that is needed to run tests

ianlongden:


ianlongden:
test.env is needed to run the tests.  Resource tests added but these are mainly just the reference ones copy and pasted.
Added tests for deletion of references and resources.

ianlongden:
Just added some more changes to the tests initialisation , so I think the tests should work for all now. (Famous last words).
After I cleaned all my docker things, the tests did not run for me either. They do now.

adamjohnwright (approved):

PR #11, opened 2021-10-21
fixing mypy errors

dougli1sqrd:
This PR will be for fixing all the mypy typing errors in the literature service repository.


backend/app/literature/crud/file_crud.py
```diff
@@ -25,7 +25,7 @@
 def create(db: Session, s3: BaseClient, parent_entity_type : str, curie: str, file_contents: str, display_name: str, content_type: str) -> str:
     filename, file_extension = os.path.splitext(display_name)
     bucket_name = 'agr-literature'
-    md5sum = hashlib.md5(file_contents).hexdigest()
+    md5sum = hashlib.md5(bytes(file_contents, "utf-8")).hexdigest()
```
adamjohnwright, line 28: I am not sure if all the files will be utf-8. some

dougli1sqrd: Yeah that's a good point. I wonder if there's a different mode other than utf-8 that would be better suited? Since the files are coming in as text (str) I assumed they'd just be text based files. If we can upload things like images too maybe the `file_contents` should actually be `bytes` or `bytearray` straight away?

backend/app/literature/schemas/reference_comment_and_correction_schemas.py
```diff
@@ -10,8 +10,8 @@
 
 
 class ReferenceCommentAndCorrectionSchemaPost(BaseModel):
-    reference_curie_from: str = None
-    reference_curie_to: str = None
+    reference_curie_from: Optional[str] = None
```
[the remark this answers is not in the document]
dougli1sqrd: ah okay, interesting. I'll think about this more.

backend/app/literature/crud/file_crud.py
adamjohnwright: I am not sure, we will have to test with different types of files. I am pretty sure I was testing with mainly random images on my computer initially. If the syntax was converting it to a string array then we need to fix it, even if it was working as is.

backend/app/literature/config.py
```diff
@@ -32,7 +32,7 @@ class GlobalConfig(BaseSettings):
     PSQL_PORT: Optional[str] = None
     PSQL_DATABASE: Optional[str] = None
 
-    RESOURCE_DESCRIPTOR_URL: str = None
+    RESOURCE_DESCRIPTOR_URL: Optional[str] = None
```
adamjohnwright, line 35: this one needs to be required. so remove the Optional and the "= None" would be correct

backend/app/literature/crud/resource_crud.py
```diff
@@ -118,20 +118,20 @@ def destroy(db: Session, curie: str):
     return None
 
 
-def patch(db: Session, curie: str, resource_update: Union[ResourceSchemaUpdate, dict]) -> dict:
+def patch(db: Session, curie: str, resource_update: ResourceSchemaUpdate) -> dict:
```
adamjohnwright, line 121: I think that Ian added the Union here when he was making the tests. I significantly prefer what you are changing it to. That being said you should make sure the tests still pass!

backend/app/main.py
```diff
@@ -2,6 +2,8 @@
 
 import argparse
 
+from typing import Dict, Any
+
```
adamjohnwright, line 6: I think it is pylint that suggests using one line per import. For example, this would be split into two lines. Just something to consider in the future.

adamjohnwright (approved):
I left a few comments. Once those are addressed I would pull it in. There are a lot of changes here and it would be good to get them in sooner rather than later. Also as a side not there is some activity in the sqlalchemy-continuum issue for adding support for sqlalchemy 1.4. I tried to make the changes and it ended up being more involved than I could do in a weekend. I am guessing they will get it done over the next while (fingers crossed). If they don't end up doing it will finish off the changes in my fork of the repo. Anyways, having these changes in there would be great for when that upgrade is available. :)

backend/app/literature/crud/resource_crud.py
ianlongden: Yes I added it. Patch fails if you pass it a ResourceSchemaUpdate but fine with a dict. This goes for most of the patch methods in all the schema files. See tests for examples.

backend/app/main.py
ianlongden: We may need a quick discussion on this as I prefer this on different lines but without the from typing on each line.
i.e.
Not
from typing import Dict
from typing import Any

but instead:-
from typing import (
    Dict,
    Any
)

As then it is easier to see the grouping of imports. But that may just be me. It would be nice if we all did the same thing but not vital or important, hence I have not said anything until now. This is especially the case for longer lists. If that list could be in alphabetical order even better, it makes finding where methods comes from easier.

adamjohnwright: Interesting. I have not seen this syntax before. I personally prefer it. I think it would be interesting to see what pylint thinks about it. I usually do the former but I would do the latter if it is generally accepted as the correct syntax in the community.

ianlongden:
Eric do you need anything from anyone to proceed with this?
Normally (well in flybase) conversations need to be resolved, this can be just saying "sorted" or "will fix later" or "I considered it but no" etc so we know what is still outstanding and what is okay.
There is the problem with the patch either not coded correctly to take the schema update or documented wrong in that in needs a dict. But from my understanding mypy does not stop anything from working it just gives warnings/errors when peppy is ran to help the developer, so no harm no foul here.

dougli1sqrd:
Hi Ian! I tried to see if I could see what was going on with the `patch` and the database object, but I was getting strange test errors. I'm having trouble running the tests, I'm not sure I have the right data in the test database? Maybe you can show me how you set that up on your machine.

But I've modified patch to handle either a `dict` or a `ResourceSchemaUpdate` so that should help. But I think this is fully ready as far as mypy goes.

ianlongden (approved):

PR #14, opened 2021-11-03
Tests cont4

ianlongden:


backend/app/literature/tests/test_001_reference.py
```diff
@@ -186,6 +188,7 @@ def test_reference_large():
     assert res['cross_references'][0]['curie'] == 'FB:FBrf0221304'
     # cross references in the db?
     xref = db.query(CrossReferenceModel).filter(CrossReferenceModel.curie == "FB:FBrf0221304").one()
+    print("BOB: xref is '{}'".format(xref))
```
gildossantos, line 191: BOB!

backend/app/literature/tests/test_002_resource.py
```diff
@@ -90,6 +90,74 @@ def test_update_Resource():
     assert res['abstract'] == '3'
 
 
+def test_resource_create_large():
+    xml = {
+        "abbreviation_synonyms": ["Jackson, Mathews, Wickens, 1996"],
+        "cross_references": [
+            {
+                "curie": "FB:FBrf09999999",
```
gildossantos, line 98: 8-digit FBrf ID is atypical - intentional?

backend/app/literature/tests/test_003_authors.py
```diff
@@ -50,14 +46,49 @@ def test_create_author():
...
+
+
+def test_show_author():
+    author = db.query(AuthorModel).filter(AuthorModel.name == "003_TCU").one()
+    res = show(db, author.author_id)
+    assert res['orcid'] == "AUT:JANE"
+
+    res = show_changesets(db, author.author_id)
+
+    print("BOB: {}".format(res[0]))
```
gildossantos, line 74: Wanna keep BOB in there?

gildossantos (changes requested):

backend/app/literature/tests/test_003_authors.py
```diff
@@ -41,7 +37,7 @@ def test_create_author():
         "first_name": "string",
         "last_name": "string",
         "name": "003_TCU",
-        "orcid": "BOB",
+        "orcid": "AUT:BOB",
```
azurebrd, line 40: For what it's worth, the format for ORCIDs in the https://github.com/alliance-genome/agr_schemas/blob/master/resourceDescriptors.yaml is 
"^ORCID:\\d{4}\\-\\d{4}\\-\\d{4}\\-(\\d{3}X|\\d{4})$"  I don't know if it's helpful or unhelpful to have a more data-like test.

backend/app/literature/tests/test_001_reference.py
ianlongden: I was not too bothered about BOB in the tests as it only is seen if the tests fail. But I will remove it as general if you see BOB in my code then it means I wanted to remove it later. (so if anyone sees BOB in my code it means it was for debugging and I intended to remove it later)

backend/app/literature/tests/test_003_authors.py
ianlongden: Killed.

backend/app/literature/tests/test_003_authors.py
ianlongden: There was no checking of format in the code so was unsure. But yes it is better to have something closer to the true values. I will update to something closer. Thanks.

backend/app/literature/tests/test_002_resource.py
ianlongden: Changed,

ianlongden: It did not matter, there is no qc on this, but yep its better to keep the test data close to what is expected. thanks.

backend/app/literature/tests/test_003_authors.py
azurebrd: Good to know, thanks @ianlongden .  Yeah, there isn't any checking of format on the data coming in either, mostly it comes from PubMed, but if a DQM submitted something in the wrong format, it would go in that way.  Same with the curation UI as it currently stands.  Well, maybe the FMS validator prevents it going in if it's not the format from the https://github.com/alliance-genome/agr_schemas/blob/master/resourceDescriptors.yaml ?  I don't know how the validator works.  In the future we could add more restrictions.  The API example for ORCID also isn't explicit, just has the  ORCID prefix.  In general this data is not too important though, at least it won't be until we start trying to connect authors to Persons.

azurebrd (approved):
It all looks good to me.  I started adding comments as I had some thoughts, but everything I didn't comment on also looks good.

About https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_005_cross_ref.py
    # Causes a crash?? recursive error?
    # res = show(db, "XREF:anoth")
    # assert res['curie'] == "XREF:anoth"
    # assert res['resource_curie'] == 'AGR:AGR-Resource-0000000001'
I don't really know, but at some point we were talking with Adam about restricting cross_references to only have one entry per prefix type for a given reference or resource.  I didn't think he implemented that, but if he did then it might explain why .... ah, nope, you're connecting XREF:123456 to a reference and XREF:anoth to a resource.  I don't know why that would crash.  I'm leaving this in this message because depending on what happens with Steven and some FB references having multiple ISBNs in xref, we might need to  move ISBNs out of xref, creating new fields, or allow xrefs to have multiple prefix types per reference (I think that's how it is now), or tag xrefs to have some kind of 'type' field to differentiate between different types of ISBN (we'll have to talk to Ceri about that down the line when we're doing FB updates).  Separately and relatedly right now the API generates URLs always from the resourceDescriptors.yaml and when there are no pages it uses the default URL, which doesn't work for FB resources, because there aren't FB resource pages, so we shouldn't link to them, but the API generates them.  For that we'll either have to suppress resource URLs at the API, or add another field to xrefs to tag whether they should generate a url or not, or leave it as is and let the UI filter that out.  Just letting you know those changes might be coming up at some point.

https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_006_ref_cor_com.py
The comments all say "# ref_cur_to missing"  instead of also ref_cur_from and reference_comment_and_correction_type

https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_007_notes.py
No one's using notes, we haven't really talked about how to deal with them, Adam just created the table and endpoints, so this seems good

https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_008_mesh_detail.py
This data only comes from pubmed and is read-only for the UI.  When we eventually deal with PubMed updates, we'll have to be able to add / remove / edit / obsolete terms (depending on how we decide to handle delete vs obsolete, probably the way we handle mod_reference_types)

https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_009_mod_ref_type.py
I thought reference_type was a controlled vocabulary for the list of MODs, but it looks like it's just a required field.  Up to you whether you'd want to make it a CV in the database/API.  It's CV in the UI, so it should be fine as-is.  More of an fyi about the example data (string1 and string3)

https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_010_ratt.py
https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_011_rmtt.py
Adam just put these together towards the end, but we never fleshed out how they should work, so I don't know if the tables's structure and API even make sense.  I'm not sure handling this is in the backlog, but I'm sure at some point we'll have to discuss this with Ceri (and possibly Textpresso/Ontomate)

https://github.com/alliance-genome/agr_literature_service/blob/tests_cont4/backend/app/literature/tests/test_012_resource_descriptor.py
I didn't know he was doing that, but it makes sense since he's generating URLs from the resourceDescriptors.yaml

azurebrd:
I see more tests were added to Notes since I started looking at it.  That's probably good, although it's not being used as we don't have curator specs on how they should get used, and it's for curators to manually enter stuff (or potentially programmatically add messages for curators to see ?)   It's not been discussed, so probably not worth spending a lot of time on.

For that matter, the current people and person stuff are completely made up.  We most likely won't work on Persons until after NLP, and at that point we'll need a discussion between WB, ZFIN, and two non-Alliance MODs about how what is needed.  It's best to pretend those tables don't exist, or even get rid of them (if you don't want to write tests for them / make it seem like coverage is low because they don't have tests)

ianlongden:
Thanks @azurebrd and @gildossantos  it was a big one . Just testing what is currently happening really. Wether it makes sense or not I was not sure but at least it is tested and we know what it does do. Coverage up to 90% now :-) Also found some bugs along the way some I fixed some I just added notes. thanks again.

azurebrd:
Awesome, thanks @ianlongden

PR #16, opened 2021-11-05
Allow ISBN to have no pages, as they should, to prevent a report ment…

azurebrd:
…ioning that

Sorry @gildossantos I didn't catch that having ISBNs was generating an entry in the report files because they don't have a "pages" attribute.  They shouldn't according to the https://github.com/alliance-genome/agr_schemas/blob/master/resourceDescriptors.yaml and this update should prevent the message in future FB_main reports.


gildossantos (approved):

azurebrd:
Thanks @gildossantos

PR #20, opened 2021-11-09
manually chose references with specific issues into inputs/sample.jso…

azurebrd:
…n  generate_dqm_json_test_set.py now has option to generate sample reference set based on sample.json


ianlongden (approved):
Looks sensible is all I can say. Like the hard coded yaml file of ids, we can add to these easily as we go.

azurebrd:
Cool, thanks Ian

PR #21, opened 2021-11-09
Fix for alembic open_access column

ianlongden:


azurebrd (approved):
I don't know alembic, but looks good

PR #23, opened 2021-11-10
Added readme for testing sample data load and update.  Renamed sample…

azurebrd:
….json to sample_dqm_load.json and created parallel sample_dqm_update.json  Renamed sample_reference_populate.sh to sample_reference_populate_load.sh and created parallel sample_reference_populate_update.sh to compare and update and run the create pipeline on new mod references.  generate_dqm_json_test_set.py takes command line inputs to work off of different input files and output directories.  Cleaned up sort_dqm_json_reference_updates.py and is in live changes state.  Added try to parse_dqm_json_reference.py for processing single mods.  Cleaned up commented out code from parse_dqm_json_resource.py


ianlongden (approved):
Not got my head around all this yet, but do not see anything jumping out as wrong. So happy to approve. I sent an email out about future testing but this would still work if we decide to go that way. I will need to run things a few time I think to see exactly what is happening but after that I hope to be more useful with these PR's.

azurebrd:
Thanks @ianlongden , I'll look at the email and reply

PR #25, opened 2021-11-11
adding github action to run flake8

dougli1sqrd:
This adds a "local" non docker flake8 target to makefile. This should be okay since we're just looking at the source/static analysis and not connecting it to any actual outside world. This will make it easier to put into a GH action.

This action is marked `continue-on-error`, so that failing flake8 doesn't stop a merge, ie it's optional. I think that's something we discussed?

### DIFF
--- a/.github/workflows/flake8.yaml
+++ b/.github/workflows/flake8.yaml
@@ -1,0 +1,21 @@
+name: flake8
+on: [push]
+jobs:
+  run-make:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-python@v2
+        name: setup python
+        with:
+          python-version: 3.9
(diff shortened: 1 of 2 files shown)


dougli1sqrd:
This does mean that currently the PR will just happily will be green even if we have lots of flake8 errors. I don't know if there are configs for flake8 or we can maybe set a threshold on the errors? At the least it now is runnable. On my branch it seems to have completed successfully, showing that the `continue-on-error` works.

azurebrd (approved):

ianlongden:
Not sure why the flake8 tests did not run here as the code should have triggered that?
I was hoping to see the output so we can decide on wether to do a quick clean up and make it have to pass.
I think we decided flake8 had to pass (mypy was the one that was non fatal).

dougli1sqrd:
@ianlongden they ran on my branch: https://github.com/dougli1sqrd/agr_literature_service/runs/4183314956?check_suite_focus=true

PR #26, opened 2021-11-12
first try at functional tests in functional_tests.py split up some co…

azurebrd:
…de in generate_dqm_json_test_set.py to import here


azurebrd:
@ianlongden I don't know the proper way to write tests in python.  About a year ago I wrote something with unittest and unittest.mock to work with coveralls, but I wasn't sure what I was doing there and don't recall it well.  I see you're using pytest, and that's not installed on the dev server.  functional_tests.py is just logging 'Success' or 'Failure' after each test.  
Separately, it takes 3 minutes to query the 34 references from the database, even though it only takes 22 seconds to load them into the database, so something's fishy.

azurebrd:
@ianlongden fyi: in figuring out the recursive pubmed calls so that they can happen or not happen outside testing or inside testing, I'm adding extra parameters to existing functions. I hope it doesn't make it hard to merge unit tests later. If you think it might, maybe we should merge this and I'll keep updating the tests I'm working on.

azurebrd:
@ianlongden static sample dqm and pubmed xml files are in.  pytest runs inside docker manually, but the tests are commented out while the database isn't working.

ianlongden:
Cool, so lets see if we can sort out the database today. This could be a good time to change from using a request to doing direct calls (i.e. what the router would call) as that will be faster and may get around the permissions problems.

azurebrd:
@ianlongden Given where testing is at now on the live server, I'm going to work on the functionality of the tests outside docker without pytest.  Have them use the test database off the API, check that using the static XML / dqm files give correct results for exact title matches and so on.  Then work on a file for updates and test that.  I'll add 'assert' checks as appropriate but we'll have to figure out how to get pytest outside docker, or how to run tests on the live server inside docker.  Then we can work on the direct calls after that.

ianlongden (approved):

ianlongden (approved):

azurebrd:
Thanks @ianlongden

azurebrd:
fyi: keeping the branch since I'll keep making changes to it for the test tickets.

PR #27, opened 2021-11-14
test fixed

ianlongden:


azurebrd (approved):
Very minor thing, on backend/app/literature/tests/test_010_ratt.py new line has "" for value and '' for field.  It's similarly inconsistent on the xml dict definition.  Also in other places in other files, so it really doesn't matter.

PR #33, opened 2021-11-15
Pr unittest

ianlongden:
This is now ready for PR.
Bad tests produce errors and good tests are good, so all is right in the world.


azurebrd (approved):
Ah, didn't realize  src/xml_processing/functional_tests.py was still in there.  It's been deprecated by tests/test_functional.py   Do you still need to rename that file, or is that file okay as-is ?

PR #34, opened 2021-11-15
load tests now check for explicit values from the github pubmed_xml/

azurebrd:


azurebrd:
commit says "tests/README.md" but it's not in "tests/"  it's in "src/xml_processing/README.md"

ianlongden (approved):
Params should be explained OR mypy added so we know what the params are.
Also remove commented out code it is in a repo now so if we need to back track etc then we can use that. Nothing serious so approving, just something to bear in mind in the future.

azurebrd:
Cool, thanks @ianlongden .  When you say Params should be explained, do you mean in the comment section at the start of each function ?  I've not used doctstrings before (I think Paulo added those comments for functions), so I'm not sure whether there's an expected pattern for explanations.
Sounds good about removing commented out code.

PR #38, opened 2021-11-22
Refactoring docs

nuin:
I recreated the branch and refactored some files that @azurebrd mentioned. There's a lot of work required in many places, mostly on `manage.py` in order to make it a full fledged manager script.

`xml_to_json` needs a lot of work on a major function that is key in the code. 

Documentation also needs work, but a good skeleton is there to guide future work.

More manual and automated tests are required.


azurebrd (approved):
Hey @nuin , that's great =)
It looks like a lot of files were copied into code/lib/ and refactored, so it's hard to tell what's different, just that they do a lot of similar things, but also different things.  If they work well for you, I'm good with it.
generate_pubmed_nlm_resource.py got moved instead of copied though ?  I haven't tried running anything, but those changes look fine.

azurebrd:
@nuin I don't know where we stand with the tests, it looks like it will allow merging even though flake8 and mypy don't pass.  Probably best to talk to Ian and Eric before merging, so future changes that fail those 2 are clearer to understand what's failing.

azurebrd:
@nuin In retrospect, with generate_pubmed_nlm_resource.py  not being in src/xml_processing/ anymore, I probably shouldn't have approved it.  Maybe restore the previous version of that file that was there ?

ianlongden:
So the failed test do not stop the merge!!! Ahh! Will look into this.
So for now I think the approver should only approve when the tests all pass.

nuin:
@azurebrd I will work on the `manager.py` script and I will restore the script that is missing, no issues. I will patch the branch and submit again.

azurebrd:
@ianlongden Makes sense, don't review until the tests finish.  Don't approve until the tests pass.

@nuin Great, thanks (restore script, work on manage.py)

PR #42, opened 2021-11-22
Refactoring docs

nuin:
Third time is a charm ...


nuin:
Same as previous PR.

azurebrd (changes requested):
generate_pubmed_nlm_resource.py  has been renamed again instead of copied, but otherwise looks good.

nuin:
strange, I will double check. Is it OK to add the version from `main`?

azurebrd:
@nuin Yeah, the one in main should be good

azurebrd (approved):

PR #45, opened 2021-12-14
fix *crud fro cross_reference, editor, mesh_detail, reference

FB-ping:
I fixed *crud patch bug for cross_reference, reference, editor, mesh_detail, still need to fix for mod_reference_type. There is no data for file, note, people (same as person ?), reference_automated_term_tags, reference_manul_term_tags, so I have not tested this yet.

### DIFF
--- a/backend/app/literature/crud/editor_crud.py
+++ b/backend/app/literature/crud/editor_crud.py
@@ -50,7 +50,7 @@ def destroy(db: Session, editor_id: int) -> None:
 
 
 def patch(db: Session, editor_id: int, editor_update: EditorSchemaCreate) -> dict:
-
+    editor_data = jsonable_encoder(editor_update)
     editor_db_obj = db.query(EditorModel).filter(EditorModel.editor_id == editor_id).first()
     if not editor_db_obj:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND,
@@ -58,7 +58,7 @@ def patch(db: Session, editor_id: int, editor_update: EditorSchemaCreate) -> dic
     res_ref = stripout(db, editor_update, non_fatal=True)
     add(res_ref, editor_db_obj)
 
-    for field, value in editor_update.dict().items():
+    for field, value in editor_data.items():
         if field == 'orcid' and value:
             orcid = value
             cross_reference_obj = db.query(CrossReferenceModel).filter(CrossReferenceModel.curie == orcid).first()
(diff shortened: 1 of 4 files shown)


azurebrd:
@FB-ping I don't know much about this, so hopefully Adam can review it.

backend/app/literature/crud/editor_crud.py
```diff
@@ -50,15 +50,15 @@ def destroy(db: Session, editor_id: int) -> None:
...
+    editor_data = jsonable_encoder(editor_update)
     editor_db_obj = db.query(EditorModel).filter(EditorModel.editor_id == editor_id).first()
     if not editor_db_obj:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND,
                             detail=f"Editor with editor_id {editor_id} not found")
     res_ref = stripout(db, editor_update, non_fatal=True)
     add(res_ref, editor_db_obj)
 
-    for field, value in editor_update.dict().items():
+    for field, value in editor_data.items():
```
adamjohnwright, line 61: I believe @dougli1sqrd added in the .dict() when he was doing the large MyPy pull request. I thought it was not necessary. You should consult with him to make sure he agrees.

backend/app/literature/crud/mesh_detail_crud.py
```diff
@@ -43,12 +43,13 @@ def destroy(db: Session, mesh_detail_id: int) -> None:
 
 
 def patch(db: Session, mesh_detail_id: int, mesh_detail_update: MeshDetailSchemaUpdate) -> dict:
+    mesh_detail_data = jsonable_encoder(mesh_detail_update)
```
adamjohnwright, line 46: I would also check with @ianlongden  to make sure that the patch functions work with the generic update functions that he made. I am not sure if we can use his functions in this context. Or maybe we should be extending what he has already done. I think his functions are only for adding things related to reference/resource.

adamjohnwright (approved):
Everything looks good as long as @ianlongden and @dougli1sqrd agree with the changes

backend/app/literature/crud/editor_crud.py
dougli1sqrd: If mypy checks out (which it seems like it does) and this runs just fine then this is fine by me. I think I was previously getting errors that items() is not a function on that type. But this seems fine by me!

dougli1sqrd (approved):

ianlongden (approved):

PR #46, opened 2022-02-02
Pgsync schema with child nodes and wildcard field for cross ref curie

valearna:
This PR adds authors, resources and cross_references to the index by defining relationships in the pgsync schema. In addition, the field cross_references.curie is set as an elasticsearch wildcard field (https://www.elastic.co/guide/en/elasticsearch/reference/7.16/keyword.html) through the 'mapping' field in the schema.

This is a sample query that can be run on the resulting index:
curl -H 'Content-Type: application/json' -XPOST "http://localhost:9200/references_index/_search" -d '{"query": {"bool": {"must": {"wildcard": {"cross_references.curie": "PMID\\:12345*"}}, "filter": {"match": {"cross_references.is_obsolete": "false"}}}}}' | python -m json.tool


azurebrd (approved):
Thanks @valearna !

PR #48, opened 2022-02-04
changed date_published field from 'date' to 'text'

valearna:
Some values in the DB for the field references.date_published cannot be converted to date, but elasticsearch was automatically setting the field type to date and was failing on these entries. Setting the field type to text in pgsync_schema to solve the issue


azurebrd (approved):
Sorry @valearna , was on a long call until now.

PR #50, opened 2022-02-08
feat(pgsync): added custom field configuration for A-team

valearna:
custom config for specific queries that the A-team needs on curies. Other fields might need custom config (to be discussed with A-team)


azurebrd (approved):
That's great !

PR #52, opened 2022-02-09
Python goodness - second try

nuin:
Added `poetry`
Ran `isort` on all python files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ pytz = "^2021.3"
 databases = "^0.5.5"
 fastapi-auth0 = "^0.3.0"
 isort = "^5.10.1"
+Sphinx = "^4.4.0"
 
 [tool.poetry.dev-dependencies]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,41 @@
+[tool.poetry]
+name = "agr_literature_service"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
+
+[tool.poetry.dependencies]
+python = "^3.9"
+requests = "^2.27.1"
+pytest = "^7.0.0"
+black = "^22.1.0"
+mypy = "^0.931"
+flake8 = "^4.0.1"
+SQLAlchemy = "^1.4.31"
+alembic = "^1.7.6"
+fastapi = "^0.73.0"
+PyYAML = "^6.0"
+fastapi-health = "^0.4.0"
+uvicorn = "^0.17.4"
+SQLAlchemy-Continuum = "^1.3.12"
+pydot = "^1.4.2"
+boto = "^2.49.0"
+bcrypt = "^3.2.0"
+python-jose = "^3.3.0"
+python-multipart = "^0.0.5"
+psycopg2-binary = "^2.9.3"
+asyncio = "^3.4.3"
+asyncpg = "^0.25.0"
+passlib = "^1.7.4"
+python-dotenv = "^0.19.2"
+SQLAlchemy-Utils = "^0.38.2"
+pytz = "^2021.3"
+databases = "^0.5.5"
+fastapi-auth0 = "^0.3.0"
+isort = "^5.10.1"
+
+[tool.poetry.dev-dependencies]
+
+[build-system]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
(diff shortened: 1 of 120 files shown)


pyproject.toml
```diff
@@ -0,0 +1,41 @@
+[tool.poetry]
```
adamjohnwright, line 1: I was using requirements.txt because it worked well inside of Docker. If this is supposed to replace the requirements file then make sure this works when building the Docker container.

azurebrd (commented):
I don't really get it, is this mostly formatting and relatively less additions of new things ?  I don't know what it's supposed to do, or if things still run properly after the changes.  I know @ianlongden is making changes to the repo, and I imagine this won't cause conflicts, but I don't know.  I guess, if other people are okay with it, that's fine with me.

adamjohnwright (approved):
organizing the imports alphabetically was on my list of things I always wanted to get done. This tool that does it for you is great. I was importing one dependency on each line because that is what I read to be best practice when I started this. I think that the gains from using this tool to check for depricated dependencies etc well outweigh the benefits from splitting out the dependencies. Also removing the repetition is a lot more readable for files where we have a lot to import. Looks good... just make sure the TOML dependency file works with Docker and the group agrees with using it!

pyproject.toml
nuin: It is possible to generate a requirements.txt from poetry and use it, or use poetry directly on docker

nuin: Also, I haven't removed the requirements. poetry would work well for local development and some testing and running thinks locally.

ianlongden:
Looking through this now, I will add a new comment each time something springs to mind.
Firstly:-
Is there a way to specify the number of methods per import. Presently it is 2 but I would prefer 1.
So the new import is:-

from literature.models.resource_descriptor_models import (
     ResourceDescriptorModel, ResourceDescriptorPageModel)

But I would prefer:-
from literature.models.resource_descriptor_models import (
     ResourceDescriptorModel,
     ResourceDescriptorPageModel)

I find it much easier to find what I am looking for then when they are all in a vertical line. Not a deal breaker and could be just my preference but ........

ianlongden:
Otherwise all looks good. I was scared it was going to alter most of the formatting. But it is just the imports so fine, I must have skipped that bit in the chat.

nuin:
@ianlongden We can define a preferred method

https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html

ianlongden:
Cool I would like to opt for "3 - Vertical Hanging Indent" But we may need to check with others as that could be just my preference and others hate it.

azurebrd:
> Otherwise all looks good. I was scared it was going to alter most of the formatting. But it is just the imports so fine, I must have skipped that bit in the chat.

The  other PR, I think with 'black', alters some of the code formatting

nuin:
if anyone can approve this I would appreciate

azurebrd (approved):
After the standup, I think everyone was okay with this PR.  Thanks @nuin

ianlongden (commented):
I would like to change the format to type 3 id no one objects. Do we need to do that before or after the merge f people agree on this.

azurebrd:
*shrug* for the imports, I don't have a real preference on the formatting

dougli1sqrd:
I feel pretty dubious about this to be honest.

I'll go with it if everyone else wants this though.

1. I find the general idea of enforcing a strict formatting style to be mostly unnecessary. Within a certain set of reasonable guidelines I think different ways of say writing an import or what quotation mark to use is very surface level. For example, if we're using the `from x import (a, b, c)` with a b and c being on different lines, why prefer that over three separate lines of `from x import a` etc? I just don't really understand what this is solving.
2. Regards to poetry, maybe I'm just "old man yelling at cloud meme" here (https://i.kym-cdn.com/entries/icons/original/000/019/304/old.jpg) but if poetry was used would the project now _depend_ on poetry for its dependencies? Like we would have to use it? What does poetry get us over pip and virtual environment and setup.py and requirements? I just feel uncomfortable using sort of non-standard tools that seem to be very similar to standard practices that already exist. I see we can generate a requirements.txt, but then we might as well use requirements.txt, right?

Thanks!

azurebrd:
@dougli1sqrd good points

nuin:
So, @dougli1sqrd for `poetry` you can use it if you want or not, it's up to you.

Regarding `isort` we have to believe that the code is well formatted and follow specs because some dat we won't be here to work on it and someone else will.

@ianlongden I will merge this, setup the configuration and then reformat. I will try to add some other changes too, and merge whatever you changed

PR #55, opened 2022-02-09
SCRUM-1174

ianlongden:
This seems to work now. To run the functional tests we can now just type
make run-functest

NOTES:
1) We can use an alternative api server now by setting an env of API_SERVER but this is not used presently. as the localhost in the docker network is sufficient.

2) to view the log files from the tests you can do something like:-
docker exec -it \`docker ps --no-trunc -aqf name=agr-literature-app-test\` ls /logs
or just go into the container and rummage around
docker exec -it \`docker ps --no-trunc -aqf name=agr-literature-app-test\` /bin/bash

In the future we may need to (But I am time limited at the moment) :-
1) Presently I am unsure if the tests will fail with exit code if there is a problem or wether we just need to eyeball the results.

2) Might be easier to move most of the make command in run-functest into a shell script that way we do not need to do a more of the results and can also examine the output file and just report any errors.


azurebrd:
I've updated the documentation at https://agr-jira.atlassian.net/wiki/spaces/BLUE/pages/533856257/Dev+Server+and+Literature+Stack with comments from this PR.

ianlongden:
> That's great, thanks @ianlongden !
> 
> When doing the run-functest and it's populating a load and update, what's the directory it's on (to know where the load.out and update.out go) Do they get routed to the /logs/ directory ?
These go to wherever you ran it from.


> It might be nice if the echos into the /logs/ in sample_reference_populate_load.sh were less "Bob" :)
I will search for Bob and eradicate him wherever I find him.

> 
> There's a print on line 46 of functional_tests.py I'm not sure if that's on purpose or left over from testing.
Yep I missed that one. I will remove it when I deal with "Bob".
> 
> I've not worked with the backend side of the repo, or with docker much. Where does the database name get set and how does the API know which database to connect to ? In the main docker-compose.yaml I see postgres+pgadmin volumes that make sense, but don't see a database. I'm mostly curious about how it works, but also concerned if we run that on the dev server that it could affect a running dev database. (assuming the database is called 'literature' that's fine, it exists and it's not being used)

By setting the environment ENV_STATE to test it uses the file test.env to get all its configuration data.
Here is where the database name is set. 

For the server it is always localhost presently BUT if you set the API_SERVER environment variable before running it will use that server name. Currently not used just there for if we need it.

ianlongden:
Dev database should be safe no matter what happens as when you call it, the database used has to be on the docker network created. 
The docker network created does not have access to anything outside apart from the volumes. 
The volumes are named differently here and have -test on the end.
It would take some deliberate changes to mess everything up here.

azurebrd:
Oh, awesome, thanks @ianlongden 
Good to know it can only access the database in the docker network, the docker network cannot access outside volumes, the volumes are named differently, ENV_STATE pointto test.env for the DB name, API_SERVER can override localhost, logs go where the code's run, Bob-eradication as appropriate :)

PR #57, opened 2022-02-13
New refactoring

nuin:
Mostly files from the XML pipeline

Added a call graph to understand module/function dependencies

Needs more work on really complex and long functions, documentation, simplification, etc


azurebrd (commented):
That's a lot to look through.  Is there anything changed with the logic ?  scripts have moved to other directories, so I can't tell what's really changed.  I'd say, you should really run everything before and after, and check that the output is the same.  If it's the same, then okay.  If not, I don't know, I guess we should all decide that you're taking over the xml processing completely.  Like, I haven't done stuff like     
resource_abbreviations = (open(resource_abbreviation_not_found_filename).read().splitlines() )   So I don't really know.

I'm still not fond of things in () split into different lines, like logger.info(stuff) becoming three lines, when it's basically a print statement.

nuin:
No changes in logic, probably did a couple of minor changes that might need checking, and as it's in a different directory there's no "takeover" as nothing actually changed.

The idea here is to make the code less complex, more Python-like and easy to understand. There are some really large functions that might need to be broken down and refactored. Maybe we can find efficiencies here and there.

As we planned, I can do more stuff after going over the code with you, then I can suggest larger changes.

azurebrd (commented):
[src/xml_processing/code/lib/get_pubmed_xml.py](https://github.com/alliance-genome/agr_literature_service/pull/57/files#diff-baa2bcce81ee9e70cfb0b0e983d0d18626b38d1923ffc89582159b6e648dff9a)   has a bunch of stuff uncommented that are pipenv commands and text comments, so that's probably not right.

Okay, so if I understood, there's a bunch of new files in the lib/ directory, and for now I can just ignore what's happening there.  Then we'll talk on Tuesday to go over stuff, and at some point you'll run the old scripts then the new scripts, and compare their outputs before replacing the old scripts ?

Also see changes to the backend/ that seem about styling, so that seems fine with me.

nuin:
For [src/xml_processing/code/lib/get_pubmed_xml.py](https://github.com/alliance-genome/agr_literature_service/pull/57/files#diff-baa2bcce81ee9e70cfb0b0e983d0d18626b38d1923ffc89582159b6e648dff9a)

basically I added all comments after the import to a DOCSTRING on the top of the file, just to organize. 


Yes, the idea is that to change anything, I would need to understand what is going on, and I don't want to do anything without you checking first. Everything so far were style changes and some simplification of the code in different places that don't interfere with the original as they are in different dir.

And anything on crud will be documentation and minimal style changes, like consistency with double quotes instead with a mix double and single.

azurebrd (approved):
Awesome, thanks @nuin

PR #60, opened 2022-02-14
remove printing of entry and change bob to test

ianlongden:
The "Bob" in the tests I have not changed as these are just tests and I do not have time to change these presently.
I can do these when I return to sprints but this seems trivial for now.


azurebrd (approved):
Cool, thanks

PR #63, opened 2022-02-15
Pr functest

ianlongden:


azurebrd:
Hi @ianlongden .  What's not working ?  It would probably help to get feedback from @adamjohnwright if it's about Okta

ianlongden:
Okay I think I got it now, no idea why it was not working earlier!!!!
Github actions now requires the functions test to pas. which they do. :-)

azurebrd (commented):
woohoo !  
I'm kinda surprised that the AWS keys need to be removed.  I'll wait for @valearna and @nuin to comment, since they know better about that.

azurebrd (commented):
Thanks @ianlongden   By de-bobbifying I meant the echos of 'asdfs' (or whatever it was), I didn't mind the PGADMIN bob
I think Valerio's still talking with you about this, so just a comment here.

nuin (approved):

PR #64, opened 2022-02-20
Start of speed testing and pydantic config changes

ianlongden:
NOTE: there is no 'build' state for ENV_STATE now. It is either 'prod' or 'test'. We can add build back in if needed. pydantic now looks for a relative path of the module rather than just in the directory to get the env file. This works a lot better now as we can start the app, alembic etc from wherever.
Hopefully everything else makes sense.


azurebrd (commented):
There's a section commented out on line 542 in sort_dqm_json_reference_updates.py that would be used if curators want reports on resources changing, and the agr_url / url_ref_curie_prefix is only used if they want that report, so it seems okay to comment out those lines (at 475 and 486 in these changes) because they're unused variables, but they're in a different section so they don't have comments about that "if curators want it" condition.  Maybe @nuin would want to make it all explicit but have flags for which reports to generate, and then everything would get uncommented ?  (I'm concerned that commenting them out where they are without context would lead to them getting removed in the future and then the commented out section wouldn't work if uncommented).  I don't know what best practices are.

azurebrd (commented):
That's great you got it running @ianlongden .  I still have the same comment about the removed comments, and maybe @nuin or @valearna have an opinion about that.

azurebrd (commented):
Also, why hardcode REG ?

ianlongden:
I could not get the REG into the docker-compose file, I tried various methods bu too no avail. As for the comments then I can put it back but as it was not documented or used I presumed it was old code. I can put it back but someone should document what it is there for or add comments to the commented out code.

azurebrd:
Good to know, thanks.  Maybe Valerio can figure out REG in https://agr-jira.atlassian.net/browse/SCRUM-1221   
About the removed commented out code, it goes with commented out code later that has explanations, and I don't know what best practices are.  I'll defer to @nuin since he's been doing the refactoring of the xml_processing

ianlongden:
Best way is to allow a flag set on running (preferable via an argument supplied to the script) to stay wether the code is ran or not usually.

azurebrd:
Cool, makes sense.

nuin (approved):

PR #65, opened 2022-02-22
Moved .env files to root dir and unified docker-compose file for main app, unit tests and functional tests

valearna:
I've tried to run unit tests and functional tests and the output seems ok to me, but please @ianlongden take a look at the changes and make sure everything is ok before merging this PR. I haven't merged the docker-compose files yet but I think that we can set COMPOSE_PROJECT_NAME env var (https://docs.docker.com/compose/reference/envvars/#compose-project-name) in our .env files to create different docker images for unit testing, functional testing and production without having to create additional yml files. I can work on those changes and push them to this branch


azurebrd (commented):
Thanks @valearna .  What's the difference between container_name and hostname ?  I'm wondering what the advantage of hostname is, and whether it can still be seen / stopped / affected by docker container ls (by a process-recognizable name)

valearna:
Thanks @azurebrd. `hostname` sets the hostname for the container on the local docker network whereas `contianer_name` forces the container name to be a specific string. The idea is to avoid specifying `container_name` so that the container name becomes COMPOSE_PROJECT_NAME + "_" + service_name + " _n" where n is an incremental number depending on the number of container replicas. By setting COMPOSE_PROJECT_NAME we can thus have different instances of the same image running under different component names without having to define multiple docker-compose.yml files with different container_name. The container hostname is set to be equal to the container name by default, so by specifying `hostname` for the services we can always connect to them from the local docker network using a fixed set of hostnames. Note that also docker networks are concatenated to COMPOSE_PROJECT_NAME so that containers in different networks will have the same hostname but they will be visible only to other containers in the same subnet.

valearna:
The commands to stop containers/volumes in the Makefile already work with names like `agr-literature-test_agr_literature_1`. For example, `docker ps --no-trunc -aqf name=agr-literature-test_agr_literature` matches that name too

azurebrd:
@valearna wow, that's really cool !  Is the container name generation via concatenation in any of the files changed ?  I couldn't find it, and wondering how that works.  No rush though, since you're still working on this.  Thanks.

azurebrd:
Cool, thanks !

azurebrd (approved):
That was great to look through, thanks @valearna !

ianlongden:
I presume that literature is now on a path where python automatically finds it as that is not added to the PYTHONPATH and it works. Everything else looks much neater so happy for it to be merged :-) Thanks Valerio, this should be much cleaner/clearer going forward.

ianlongden (approved):

valearna:
I set the PYTHONPATH in docker-compose, it's hardcoded though. We could transfer it to the .env files if needed. I will merge this PR as is and we can talk about other ways to improve/reorganize env vars later.

PR #66, opened 2022-02-23
Speed test SCRUM-756

ianlongden:
Looks like sql alchemy with around 500 entries might be the sweet spot for best optimisation.
1000 works but better to go low incase we get memory problems.
10,000 works but the swapping significantly slows things down so definitely out.


nuin (approved):

azurebrd (commented):
@ianlongden it's merged, so I can't review, but here's a comment.

So it seems when we update the xml updates code, we should use [sqlalchemy_call.py](https://github.com/alliance-genome/agr_literature_service/pull/66/files#diff-6fd25c95942588642ecd8711815fc125e16c3d3640946ee49ced12a441b54db3) and explicitly batch_alchemy ?  It's so little code, that's great.  Do you have a sample of a couple of references in the output ? (wondering if it's got all the authors and everything in the same format, I imagine so, since it's going through sql alchemy).   I'll try to run that on the dev server soon.

ianlongden:
Feel free to turn verbose on and try it. It would be good to know if other people can run it too.
You can comment out the ones you do not want to run and maybe just do batch size of 10 to get the fewest results.

-Ian.

azurebrd:
Cool, will do.  Thanks.

PR #69, opened 2022-03-04
SCRUM-1110  batch sqlalchemy updates seem to work as well as one-by-o…

azurebrd:
…ne through API now, but are a lot faster.  batch size is 1000, above that the dev server might freeze.  tested with sample_reference_populate_load.sh + sample_reference_populate_update.sh (using the new sort_dqm_json_reference_updates_batch_sqlalchemy.py ) + functional_tests.py


azurebrd:
@valearna @ianlongden I could use some help. 

At https://github.com/alliance-genome/agr_literature_service/runs/5418925146?check_suite_focus=true#step:6:1074 the literature module fails to load.  When running this locally I set the PYTHONPATH like 
  export PYTHONPATH=/home/azurebrd/git/batch_queries/agr_literature_service/backend/app/:/home/azurebrd/git/batch_queries/agr_literature_service/src/xml_processing/
and that solves the problem.

I don't know how to set this in Docker, or if it's even about PYTHONPATH or something else that would let docker load the modules from backend/app/    I tried to add it to the .env like
  PYTHONPATH=/usr/local/bin/src/literature/backend/app/:/usr/local/bin/src/literature/src/xml_processing/
but that doesn't seem to work.

Do either of you know ?  Thanks.

azurebrd:
https://github.com/alliance-genome/agr_literature_service/runs/5419704100?check_suite_focus=true#step:6:1072 
Adding it to docker-compose.yaml PYTHONPATH under agr_literature seems to get it to load, but cannot make connections to the database

could not connect to server: Cannot assign requested address
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?

valearna:
@azurebrd Have you tried to set PYTHONPATH in the github action file? https://github.com/alliance-genome/agr_literature_service/blob/main/.github/workflows/functest.yaml

azurebrd:
Adding to the github actions workflow has the same error.  https://github.com/alliance-genome/agr_literature_service/runs/5425498112?check_suite_focus=true#step:6:1107

I don't think it makes sense to remove it from the docker-compose.yaml  but trying it now just in case

azurebrd:
Removing from docker-compose.yaml goes back to the error of not being able to load the literature module
https://github.com/alliance-genome/agr_literature_service/runs/5425581193?check_suite_focus=true#step:6:1077

Adding it back to docker-compose

azurebrd:
@valearna I forgot we also made a couple of changes to the ports for postgres in [docker-compose.yaml](https://github.com/alliance-genome/agr_literature_service/pull/69/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9)   If you think I should undo that to avoid conflicts with your branch, let me know, but hopefully it's not a lot.  I'm not sure if it would cause github actions issues to remove them, or if it was just for my local docker run.

valearna (approved):

PR #70, opened 2022-03-05
Scrum 1221: New Docker Structure

valearna:
This PR simplifies the structure of dockerfiles and docker-compose services. These are the major changes:
- Removed hierarchy between docker images and renamed dockerfiles
- Created separate docker-compose services for API, xml_processing and test_runner
- Simplified Makefile by using docker-compose commands instead of pure docker ones

Note that make commands do not rebuild images every time they are executed but only if no image is present. `make build` or `docker-compose build` need to be run manually to rebuild images.


Makefile
```diff
@@ -8,96 +8,57 @@ endif
...
-	-docker run -i --rm \
-		--network=agr-literature-test_agr-literature \
-	    -p 8080:8080 \
-		-e PYTHONPATH:/workdir/src/xml_processing/ \
-	    -v ${PWD}:/workdir \
-		${REG}/agr_literature_dev:${TAG} \
-		./run_tests.sh > pytest.out
-	docker-compose --env-file .env.test down
-    #doing here after shutdown of database 
+	-docker-compose --env-file .env.test run -v ${PWD}:/workdir test_runner ./run_tests.sh > pytest.out
```
azurebrd, line 46: There's a leading - here, but maybe that's supposed to be there ?

docker-compose.yaml
```diff
@@ -97,36 +108,141 @@ services:
...
+      dockerfile: docker/xml_processing.dockerfile
+      args:
+        REG: ${REG}
+        ALLIANCE_RELEASE: ${ALLIANCE_RELEASE}
+    hostname: literature_xml_processing
     environment:
       API_PORT: "${API_PORT}"
       PYTHONPATH: "/usr/local/bin/src/literature/src/xml_processing:/usr/local/bin/src/literature/backend/app"
-      XML_PATH: "${XML_PATH}"
-      PSQL_DATABASE: "${PSQL_DATABASE}"
```
azurebrd, line 117: I'm wondering why PSQL_DATABASE is not needed in xml_processing, but is needed in test_runner   Right now the batch sqlalchemy is doing a connection to the database, and that's only in the functional tests in test_runner, but eventually I would think Paulo's refactor would have the xml_processing also use the batch sqlalchemy queries (the code that does it, even for the functional tests, is part of that section of code)

azurebrd (dismissed):
That's great, thanks Valerio !
I left a couple of comments, one minor one, one wondering about PSQL_DATABASE   I imagine it's all working, so approving, but not merging in case you change things based on the comments (not saying you should)

Makefile
valearna: the leading - suppresses errors and it's needed here.

docker-compose.yaml
valearna: I haven't really tested the xml_processing service and I don't know which variables will be needed, but I will add them based on your comment

Makefile
azurebrd: Good to know, thanks !

docker-compose.yaml
azurebrd: Cool, thanks

azurebrd (approved):

valearna:
Thanks @azurebrd!

PR #71, opened 2022-03-07
Ignore non_pr_tests

valearna:
Added option to pytest to ignore files in non_pr_tests


azurebrd (approved):
Makes sense and seems good, thanks @valearna

PR #73, opened 2022-03-07
pytest options moved to pyproject.toml

nuin:
Moved pytest options to pyproject.toml

Can you take a look and test @valearna?


valearna:
Tested locally. Passing the options through the config file looks much better. Thanks @nuin

valearna (approved):

PR #75, opened 2022-03-12
feat(docker-compose): removed dependencies from services that don't need them in all envs

valearna:
- some services depend on other services only for local deployment (e.g., pgsync needs postgres and elasticsearch when the whole stack is started locally but not in production, where pg and es are hosted on RDS)
- the current version of docker-compose on the dev server (which is also the default one in ubuntu 21.10) does not manage different dependencies in multiple profiles
- we decided to remove the dependencies from docker-compose.yaml and add entries in the Makefile to manage the dependencies for different profiles there


azurebrd (approved):
Thanks @valearna , what you said at stand up sounded good.

PR #76, opened 2022-03-17
new api endpoint for searching references through elasticsearch

valearna:
- added 'search' router and crud in API code
- uses elasticsearch python library to connect to es index and search using query string passes to the api
- added ELASTICSEARCH_{HOST,PORT} env variables to api and test_runner


azurebrd (approved):
Great, thanks @valearna

nuin (approved):

PR #77, opened 2022-03-23
added some changes to isort config that might fix version issue

nuin:
skipping `__init__.py` files and `database/main.py`

More can be added


azurebrd (approved):
Thanks @nuin

PR #78, opened 2022-03-23
SCRUM-1243 Take endpoint to query all cross_references, extract all c…

azurebrd:
…urie prefixes, match to MODs, generate a mod_corpus_association entry, post to API that does not exist yet, when ready update to work properly

### DIFF
--- a/src/xml_processing/process_mod_corpus_association_to_api.py
+++ b/src/xml_processing/process_mod_corpus_association_to_api.py
@@ -1,0 +1,81 @@
+
+import json
+import logging
+import logging.config
+from os import environ, path
+
+from helper_file_processing import (generate_cross_references_file, load_ref_xref)
+from helper_post_to_api import (generate_headers, get_authentication_token,
+                                process_api_request)
+
+# query database for xrefs, extra MODs, post to populate mod_corpus_association
+# python process_mod_corpus_association_to_api.py
+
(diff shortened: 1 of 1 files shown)


nuin:
I have some changes to suggest, will do on GitHub

azurebrd:
Hi @FB-ping .  This would be the simple way to post to the database one-by-one using the existing xml processing helper modules to generate the okta token and post to the API.  It downloads all the reference cross_references from the API endpoint and processes them, but since we don't yet know the final format for creating a mod_corpus_association, we'll have to update this once we do.

azurebrd:
@nuin This is a one time script that we'll run once the API for mod_corpus_association is live.  It will only run once, and we'll never use it again.  We'll probably do some modified version for the md5sum script, once that's ready to be worked on.  It's probably not worth spending a lot of time to make this really nice, but if you wanted to make a generic future-processing-to-api script, that's probably good.  (That said, feel free to make suggestions / improvements)

nuin (approved):

nuin:
Of course flake8 comes back to bite me, sorry

azurebrd (commented):
I like the changes, thanks @nuin !

nuin (approved):

azurebrd:
Thanks, @nuin !

PR #81, opened 2022-03-28
fixing mypy errors

dougli1sqrd:
Simple fixes for mypy checks in the mod_association branch


dougli1sqrd:
Oh I see some errors in mypy still, weird that locally I don't see it

azurebrd:
@dougli1sqrd The mypy errors in github actions can be fixed by merging main, which has an updated mypy.config that excludes that directory, until Paulo can come back to fix them (unless you want to also look into fixing those)
https://github.com/alliance-genome/agr_literature_service/runs/5727322914?check_suite_focus=true

azurebrd:
Yeah, Valerio and I weren't seeing those mypy errors locally.  He think we should figure out how to have the same mypy environment locally and in github actions, but we don't know how.

azurebrd:
Thanks @dougli1sqrd .  It looks good to me, but I'll defer to Pinglei and Valerio, who have actively worked on the branch.

dougli1sqrd:
infinite flake8 errors, but mypy now fixed

PR #87, opened 2022-04-08
fix(api): removed lazy load option from mod and moved it to mod_curpu…

valearna:
…s_association model

- using lazy="joined" to always load the related table into an object using a join
- modified reference_crud.py to reflect changes in the models


azurebrd (approved):
Thanks @valearna

nuin:
Pizza mezzaluna, no mezzo giorno.

PR #98, opened 2022-04-13
Scrum 1455 - fixed bug with multiple mod-corpus associations

valearna:
When multiple mod-corpus associations are present for a given PMID, the sqlalchemy query returns all the related rows as different tuples. The new code picks the row containing the association to the provided MOD if present and otherwise sets the MOD value to None. The old code was always setting the value of the last tuple in the list.


azurebrd (approved):
Thanks for catching that @valearna

PR #105, opened 2022-04-21
Scrum 1253

azurebrd:
Processing a single PMID should now update md5sum in s3.  Generalized functions.

### DIFF
--- a/.env
+++ b/.env
@@ -9,7 +9,7 @@ RESOURCE_DESCRIPTOR_URL='https://raw.githubusercontent.com/alliance-genome/agr_s
 OKTA_DOMAIN='dev-30456587.okta.com/oauth2/default'
 OKTA_API_AUDIENCE='api://default'
 
-# Envstate should either be build or prod
+# Envstate should either be build or prod or test
 ENV_STATE=build
 HOST=localhost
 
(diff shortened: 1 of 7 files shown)


azurebrd:
@valearna while trying to fix the functional tests I had to restart a devserver database and api, which I attached to the dev elasticsearch.  Maybe I shouldn't have done that ?  The devserver api on dev4006-literature-rest is now pretty slow.  If this helps debug pgsync, I just ran the functional test, wiped some tables, ran the functional test, and so on a few times.

azurebrd:
@valearna the test_runner in docker-compose uses `ENV_STATE: "${ENV_STATE}"`  Where would I look in github actions to see what values it's using ?  The xml_to_json is now downloading md5sum, updating md5sum, uploading new md5sum, but github doesn't seem to have permissions to s3.  I think that's fine, because we don't need the test runner to update our develop/ md5sum file everytime it runs, but if we wanted it to, do you know how we'd give github permission to ?  I would think it's from AWS_ACCESS_KEY_ID and  AWS_SECRET_ACCESS_KEY, but the test_runner should already have those values in github actions.

azurebrd:
Hi @valearna  I think this is all finally working properly and ready to review :)

valearna (approved):

azurebrd:
Thanks @valearna

PR #108, opened 2022-04-28
rename column for references and authors. Later part of SCRUM-1492

ianlongden:
There Amy be some print statements and a lot of coded out stuff but I will remove that when I do the table renaming.


azurebrd (approved):
That's great, thanks @ianlongden !

valearna (approved):

PR #112, opened 2022-05-02
Scrum 1491

ianlongden:
Rename authors table to author.


azurebrd (changes requested):
That's great @ianlongden 
Are the commented out lines on the author update things that alembic was trying to do, but you found were wrong ?  Did you have to add any lines manually, or was the generation of correct changes automatic ?

I see you removed the Lit_FastAPI.log that we couldn't find.  How does logging work now ? (If it works)

What is debug_functest.sh for ? (as opposed to just running functest ?)

I'm not sure the changes to src/xml_processing/parse_dqm_json_reference.py are correct (request changes is about this)  I think the use of 'crossReferences' is from dqm data, and the DQMs are still going to be sending data in the dqm format, which will have the 's' at the end.  If you're really sure that's correct and it works, then okay, but I would think it will stop finding the dqm data for those fields.

Same with parse_dqm_json_resource.py

Sort of same with post_reference_to_api.py  the camelCase stuff is coming from dqms, and the snake_case stuff is the database or API format (not sure if db or api).  So I think at least the left on remap_keys should stay camelCase.  Not sure about the right side.  For the remap_subkeys, depends on what the right side earlier worked out to be (like if mesh_terms became mesh_term in line 104, then that would be correct in line 120)

Same with post_resource_to_api.py

I'm pretty sure the src/xml_processing/pubmed_sample/*.json shouldn't change, unless the xml_to_json.py changes to remap the fields, but I think the fields are trying to match the dqm.json format.  

Same with the camelCase in sort_dqm_json_reference_updates.py

Same with the camelCase in the tests/*_sample/*.json

Same with tests/inputs/sample_dqm_updates.json

Same with camelCase xml_to_json because it's trying to generate data in the same format as dqms.

When I say camelCase, I mean the camelCase stuff, but I really mean the data that is in the dqm format.  So authors isn't camelCase looking, but it's still referring to the dqm format, so it should stay authors to match the dqm data.  

I imagine the tests pass because you're changing the source data, but the DQMs aren't going to change their source data.  Doing that would require asking them all to change their format, and testing, and also having Sierra change whatever validator handles receiving the data.   I think the best way forward it to change the data base, the xml_to_json.py back, and then change the sections that have camelCase to change all related things to keep the 's' at the end.  And hope it all works...   It might be best if we talk it over after the developer call

PR #113, opened 2022-05-02
adding unit tests for "mod" and "mod_corpus_association" 

sweng66:


azurebrd:
Hi @sweng66 
It looks like 6 tests failed https://github.com/alliance-genome/agr_literature_service/runs/6265088033?check_suite_focus=true  around line 1083  It's somehow testing `AssertionError: assert 'AGR:AGR-Reference-0000000006' == 'AGR:AGR-Reference-0000000001'`   Probably you need to run your all the tests together to check that your reference curie is the same as the one that got created for it by the tests.  @ianlongden and @valearna might have a better sense of how things should work though.

azurebrd (commented):
Nice @sweng66 !  I haven't worked on this side of the repo, so I'll defer to @ianlongden and @valearna to approve / merge.  Thanks !

sweng66:
Thanks for checking out the code @azurebrd

valearna (approved):
The tests look good @sweng66. I don't know if this is true also for other tests, but it seems that some test functions depend on previous ones in the module for data insert/update. In general I think that it would be better to have not only atomic modules, but also atomic test functions that don't depend on each other and can be run independently. Maybe this could be part of a larger tests restructuring task @azurebrd @ianlongden?

ianlongden (approved):

azurebrd:
@valearna That sounds like a good idea.  I'm not sure how the unit tests work.  I guess we'd have to create something, test, then delete before moving on to the next test ?  But then things that depend on a pre-existing object, would also have to create the pre-existing object ?  I'll defer to @ianlongden about best practices.

PR #114, opened 2022-05-02
Scrum 1491 2

ianlongden:
Ready for PR.


ianlongden:
authors, cross_references, editors, mesh_details and mod_reference_types all had table names made singular.  Still to do references, resources and users.

backend/app/literature/crud/reference_crud.py
```diff
@@ -49,30 +49,37 @@ def create(db: Session, reference: ReferenceSchemaPost): # noqa
     :return:
     """
```
azurebrd, line 51: There's still a print statement here

backend/app/literature/crud/reference_crud.py
```diff
@@ -49,30 +49,37 @@ def create(db: Session, reference: ReferenceSchemaPost): # noqa
...
 
+    logger.debug("creating reference")
+    print(reference)
```
azurebrd, line 53: There's still a print statement here

backend/app/literature/crud/reference_crud.py
```diff
@@ -113,12 +123,17 @@ def create(db: Session, reference: ReferenceSchemaPost): # noqa
...
 
+    logger.debug("add reference")
     reference_db_obj = ReferenceModel(**reference_data)
+    logger.debug("have model, save to db")
     db.add(reference_db_obj)
+    logger.debug("saved")
     db.commit()
 
     for field, value in vars(reference).items():
+        logger.debug("Porcessing mod corpus asso")
```
azurebrd, line 136: typos, but doesn't matter

src/xml_processing/tests/pmids_by_mods
```diff
@@ -1,5 +1,5 @@
 10206683	1	RGD
-10022914	3	RGD, FB, WB
```
azurebrd, line 2: I think this file shouldn't change.  It gets changed by the functional tests, but the changes shouldn't get committed.  Maybe we should add to the .gitignore ?

src/xml_processing/tests/inputs/alliance_pmids
```diff
@@ -1,6 +1,6 @@
 2
-10206683
```
azurebrd, line 2: This file also shouldn't change, but this seems like it hasn't really changed, just moved an indentifier further down

azurebrd (changes requested):
It looks really good, thanks @ianlongden !  There's a couple of files at the end with comments that they shouldn't change.  I'm pretty sure the first one is okay because the data hasn't changed, but the second one probably shouldn't change.  I could be wrong and it might be okay to have it change.  If you've tested that without the change things don't work, then keep the change.  But I think it's a change generated by the functional test that shouldn't get committed.   All the other comments are minor and fine with me.

backend/app/literature/crud/reference_crud.py
ianlongden: Done.

azurebrd (approved):
That's great, thanks @ianlongden

PR #115, opened 2022-05-04
Scrum 1516

azurebrd:
Modularized update of pubmed nlm resources to only create new resources, but not update existing ones.  This should be run before doing any new paper creation, from pubmed searches, dqm updates, pubmed paper updates, but probably not before creating one-of papers via PMID through the UI, because it would take too long.

### DIFF
--- a/src/xml_processing/generate_pubmed_nlm_resource.py
+++ b/src/xml_processing/generate_pubmed_nlm_resource.py
@@ -28,10 +28,6 @@ logging.config.fileConfig(log_file_path)
 logger = logging.getLogger('literature logger')
 
 
-base_path = environ.get('XML_PATH', "")
-storage_path = base_path + 'pubmed_resource_json/'
-
-
 def populate_nlm_info(file_data):
     """
 
@@ -91,6 +87,9 @@ def generate_json(nlm_info, upload_to_s3):
     logger.info("Generating JSON from NLM data and saving to outfile")
     json_data = json.dumps(nlm_info, indent=4, sort_keys=True)
 
+    base_path = environ.get('XML_PATH', "")
+    storage_path = base_path + 'pubmed_resource_json/'
+
     if not path.exists(storage_path):
         makedirs(storage_path)
 
@@ -132,6 +131,7 @@ def populate_from_local_file():
     :return:
     """
 
+    base_path = environ.get('XML_PATH', "")
     filename = base_path + 'J_Medline.txt'
     with open(filename) as txt_file:
         if not path.exists(filename):
--- a/src/xml_processing/post_resource_to_api.py
+++ b/src/xml_processing/post_resource_to_api.py
@@ -3,7 +3,7 @@ import argparse
 import json
 import logging
 import logging.config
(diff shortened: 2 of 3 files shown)


valearna (approved):
I don't know much about this part of the repo, but everything looks good to me

azurebrd:
Thanks @valearna .

@sweng66 Yeah, I think that's right.  When we process papers from pubmed, we connect them via NLM, and the primary key from the J_Medline is the NLM, so as long as we get the resources that have new NLMs, we can connect them to papers, otherwise we'd have to go back later to look at all the paper info to see if we can connect them to resource.  So this was more of a last minute thing so that references can get connected to resources.  We'd like to also update the resource information, and sort_dqm_json_resource_updates.py does that for DQM resources, but hasn't been tested that much and would have to be modified for pubmed resources.  It'll probably be easy to do, but Ceri hasn't prioritized us doing that yet.  The reference->resource connection is more complicated than just NLM, but essentially the NLM is the primary non-Alliance-key that we need to resolve to, so we just need to get those in.
PubMed doesn't keep papers after they delete them, so if they remove a PMID after we created it, we can't find it at PubMed later, so we lose the source of truth.  For that reason we're storing the original pubmed xml and later will store the latest pubmed xml (so that we have a record of what we used to create something for later debugging).  Similarly we don't know whether when they update the J_Medline.txt we'd lose source of truth, so we're keeping a copy of the latest file we used.  Possibly we should keep copies of all files processed, but that seemed like too much.   It's mostly for debugging though, if something's wrong with our data later it might be easier to look at the source file before trying to track through all the code.  I think it's a good point that it doesn't help us much here as with the paper XML since we'll be updating this file over and over.
I didn't know about BioPython when I wrote this.  I wrote the WormBase xml processing some 21 years ago, and found it easy to redo it in python instead of looking at new software.  We tried using some xmltodict at some point, but there was html markup in the xml values that were getting lost and the curators wanted to keep.  When Paulo was trying to refactor the xml processing, I think he tried to use BioPython, but he also started using pandas and other stuff that was making it more complicated, and the xml_to_json parser was already working well enough, so we're leaving it as-is, more for convenience and to avoid extra work.

azurebrd:
@sweng66 Yeah, we'll probably take the code from the dqm one and convert it to a module that can be used by the pubmed one.  The pubmed one so far is just a light way to get new resources before creating papers.  Depending on how fast it is to update the pubmed resources, we can make that part of the before-paper-updates script.
Yeah, I think SGD has a lot more PubMed papers than WormBase.  Even now, with Olin asking us to change the s3 storage from standard to glacierIR, aws is having issues with converting so many files.  But at WormBase, sometimes (not often) something wouldn't look right with the data for a paper, and Kimberly would see that it was correct at PubMed, and we'd find that the original PubMed file has different data, and that was the issue, which was faster than trying to debug the code.
Cool cool.  If you think importing Bio later will be good for new functionality, we totally can.

sweng66:
OK all sound good! Thanks for the details. @azurebrd

PR #121, opened 2022-05-10
updating sgd false positive pmid list

sweng66:


azurebrd (approved):
Cool.  Could you also approve my PR https://github.com/alliance-genome/agr_literature_service/pull/120 @sweng66 ?

PR #122, opened 2022-05-11
added code to check DOI in the DB and in the new paper set to avoid pulling in duplicate entries

sweng66:
1. if a new paper's DOI is already in the database and no PMID is associated with this existing one, add the PMID  to the existing paper and add the info to the log 
2. if a new paper's DOI is already in the database and is associated with a PMID, simply add the details to the log.

The log file is available at src/xml_processing/pubmed_searches/duplicate_rows.log

### DIFF
--- a/src/xml_processing/sort_dqm_json_reference_updates.py
+++ b/src/xml_processing/sort_dqm_json_reference_updates.py
@@ -197,7 +197,9 @@ def sort_dqm_references(input_path, input_mod):      # noqa: C901
     url_ref_curie_prefix = make_url_ref_curie_prefix()
 
     # download the dqm file(s) from mod(s)
-    download_dqm_json()
+    env_state = environ.get('ENV_STATE', 'build')
+    if env_state != 'test':
+        download_dqm_json()
     # to pull in new journal info from pubmed
     update_resource_pubmed_nlm()
     token = get_authentication_token()
@@ -541,7 +543,7 @@ def sort_dqm_references(input_path, input_mod):      # noqa: C901
         # env_state = environ.get('ENV_STATE', 'prod')
         # if env_state == 'build':
         env_state = environ.get('ENV_STATE', 'build')
-        if env_state != 'test':
+        if env_state == 'prod'':
             merge_md5dict = {}
             merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}
             save_s3_md5data(merge_md5dict, [mod])
--- a/src/xml_processing/sort_dqm_json_reference_updates.py
+++ b/src/xml_processing/sort_dqm_json_reference_updates.py
@@ -543,7 +543,7 @@ def sort_dqm_references(input_path, input_mod):      # noqa: C901
         # env_state = environ.get('ENV_STATE', 'prod')
         # if env_state == 'build':
         env_state = environ.get('ENV_STATE', 'build')
-        if env_state == 'prod'':
+        if env_state == 'prod':
             merge_md5dict = {}
             merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}
             save_s3_md5data(merge_md5dict, [mod])
--- a/src/xml_processing/sort_dqm_json_reference_updates.py
+++ b/src/xml_processing/sort_dqm_json_reference_updates.py
@@ -541,7 +541,7 @@ def sort_dqm_references(input_path, input_mod):      # noqa: C901
         # env_state = environ.get('ENV_STATE', 'prod')
         # if env_state == 'build':
         env_state = environ.get('ENV_STATE', 'build')
-        if env_state == 'prod':
+        if env_state != 'test':
             merge_md5dict = {}
             merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}
             save_s3_md5data(merge_md5dict, [mod])
--- a/src/xml_processing/sort_dqm_json_reference_updates.py
+++ b/src/xml_processing/sort_dqm_json_reference_updates.py
@@ -199,9 +199,10 @@ def sort_dqm_references(input_path, input_mod):      # noqa: C901
     # download the dqm file(s) from mod(s)
     env_state = environ.get('ENV_STATE', 'build')
     if env_state != 'test':
+        # download the dqm file(s) from mod(s)
         download_dqm_json()
-    # to pull in new journal info from pubmed
-    update_resource_pubmed_nlm()
+        # to pull in new journal info from pubmed
+        update_resource_pubmed_nlm()
     token = get_authentication_token()
     headers = generate_headers(token)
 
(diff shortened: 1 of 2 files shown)


sweng66:
oops forgot to run flake8 check.. will do it now

sweng66:
Do not know how to fix the following:

query_pubmed_mod_updates.py:188:1: C901 'query_mods' is too complex (26)

sweng66:
I know what to do now...

azurebrd (changes requested):
Hi @sweng66 , that's great !
A while back Adam wanted everything to go through the API.  When Blue team started people said we should query however we want, but post through the API.  I think now people are saying it's okay to post through the crud, but I wanted to confirm with @ianlongden and @valearna before approving.  I didn't know how to post through the crud, so it's great you do =)
Moving all that code into check_handle_duplicate is great, but before you were removing values from pmids_to_process, and in the function you're updating pmids but not returning it outside the function, so pmids_to_process still has all the PMIDs.  
Now that I'm looking at the code more, there's a section where
```
    for pmid in pmids_to_process:
            pmids_posted.add(pmid)
```
and that should probably happen after running check_handle_duplicate and replacing pmids_to_process with the return from that function, then pmids_posted only adds the ones that got whitelisted through check_handle_duplicate .  
@valearna should probably also comment on where the log output should go, in terms of mounting volumes.
Thanks so much, Shuai !

sweng66:
@azurebrd Thank you for your good/detailed comments.  There are a few things I want to mention:

1. Good point about posting data through API. Calling API many times can be very time consuming, but in this program, I think it should be fine since it will only need to call this API once or twice (often time it won't  need to call at all). What do you all think? 
Anyway, I am open to any suggestions.

2. The "query_mods" passes the  'pmids_to_process' list to the "check_handle_duplicate" function and later removes any duplicate entries 
directly from this pmid list ('pmids_to_process' list got updated directly) so I think it should be fine. 

To confirm this, I just tested and indeed the 'pmids_to_process' got updated. 

I can update the "check_handle_duplicate" function to return pmids if you think it will be cleaner and more readable.

3. Good catch! I moved  the following code to after running "check_handle_duplicate"   
          for pmid in pmids_to_process:
                pmids_posted.add(pmid)

4. Agreed. We need to think a good place to keep these logs. Should the program open the log_file for writing or appending? open for writing => no history; appending=>need to add datestamp and need to clean up periodically

azurebrd:
@sweng66 

1 - Cool.  Will defer to Ian and Valerio

2 - Oh !  I didn't realize python worked that way (I haven't been using it long).  So when the outer function passes it as a parameter to the inner function, it's really a pointer to the original list and removing from the list without returning it to the outer function still affects it ?  There's probably a lot of code in there that is returning stuff it doesn't need to return, if that's the case.  

3 - cool cool !

4 - I think this is the first script that is going to be run every day.  As long as it only runs once a day, maybe write but have the date be part of the filename ?  Or maybe append in case it ever runs twice.  We'll have a lot of other log files to figure out, this script is calling a bunch of other things that are also generating logs.  There's a bunch of logger.info and logger.debug around, we probably want to redirect those into the outer volume.  We need clarity from @cerivs and curators about which messages they want for reports, and which logs are just for us to debug.

sweng66:
Also, I am not sure I understand this correctly since I haven't looked through things under /backend/app/.  Seems like there are two kinds of  "pages": one is ["reference"]; another one is "resource".  see "add" function in backend/app/literature/crud/reference_resource.py.

sweng66:
Good idea to append the datestamp to the log_file names. Just we need to clean up
the log files weekly or monthly?

sweng66:
2. yes. but sometimes we don't want to alter the original list so return a new list will be a good idea.

sweng66:
4. open file with datestamp for appending sounds good to me.

azurebrd:
I don't see 'pages' under add.  There used to be 'pages' field in the reference table, which has been replaced with page_range, making things less confusing.  
This file https://github.com/alliance-genome/agr_schemas/blob/master/resourceDescriptors.yaml generates the mapping between a curie prefix and the URL it should correspond to, so for example, for SGD, if you had a curie and wanted to construct a URL, you would use `http://www.yeastgenome.org/[%s]`  But if you knew that the datatype you had was a reference, you'd want to use 
`  - name: reference
    url: https://www.yeastgenome.org/reference/[%s]`
so you'd give it the 'pages' attribute 'reference' so that the URL constructor would use that url instead of the default_url.
PMID doesn't have a 'reference' attribute, so I don't think it makes sense to add it to PMID connections.
DOI does have a 'reference' attribute, but I don't understand why it does, since it's the same as the default_url, so I think it would make more sense to not have the reference attribute and use the default.

Here's some example data from the API
https://dev4002-literature-rest.alliancegenome.org/reference/AGR:AGR-Reference-0000641424
with pages for WBPaper but not for DOI nor PMID (under cross_references)

Logs :  I don't know how often we'd need to clean up, I don't know how big the files will be, or how soon we might be able to move to logstash.  I think ideally we'd keep all the logs so that we can track what went wrong, but maybe I'm thinking that because WB never had that much data.  Still, the logs for creating all the references at ABC are not that big.

2 - Well, good to know that it's possible to just replace like a list pointer =)  I'll defer to you on how we should do it.

4 - cool cool (filename with datestamp).  Hopefully we hear from Ian and Valerio about these things tomorrow.

sweng66:
Thanks for the info about "pages" parameter. I have removed this parameter and the program is still working in adding a PMID to an existing paper. =) @azurebrd

azurebrd:
Valerio says update though the crud is good.
Logs in the filesystem for now, logstash later.

azurebrd (approved):

azurebrd:
Thanks @sweng66 !

src/xml_processing/sort_dqm_json_reference_updates.py
```diff
@@ -541,7 +541,7 @@ def sort_dqm_references(input_path, input_mod):      # noqa: C901
         # env_state = environ.get('ENV_STATE', 'prod')
         # if env_state == 'build':
         env_state = environ.get('ENV_STATE', 'build')
-        if env_state == 'prod':
+        if env_state != 'test':
```
azurebrd, line 544: no no, this was correct before

azurebrd:
@sweng66 The last change you made was correct before the change

I meant that in https://github.com/alliance-genome/agr_literature_service/commit/3c14a7b5c72c42ea6da7cadf5b37cc537a5994b6 line 200 has download_dqm_json()  but that should only happen if env_state is not test, otherwise the functional tests take an extra 11 minutes downloading the DQM files.  we only need that when env_state is develop/build/prod

azurebrd (dismissed):
Fantastic, thanks @sweng66 !

azurebrd (approved):

azurebrd (approved):
Thanks @sweng66 !

azurebrd:
That's great, I see the functional test is back to less than 4 minutes.  I've merged it but haven't removed the branch in case you're still doing stuff with it.

sweng66:
Thanks @azurebrd

PR #127, opened 2022-05-25
adding language field

sweng66:


ianlongden (approved):

sweng66:
Thanks! Ian.

azurebrd:
Oh, thanks for catching that @sweng66 .  I thought the data was being capture, but I guess it was only getting captured if the Language tag was withing the OtherAbstract tag ?   These are the original specs for what we were supposed to capture from @cerivs https://docs.google.com/document/d/1SmDxCnvSQ4HjTLL31fI2hgCdcjCcFC2DgaHCbWt1BY0/edit starting on page 9.  I'll write to her to clarify / confirm that we should be capturing that Language tag you're capturing.

sweng66:
Good to know! Thanks for the clarification.

PR #128, opened 2022-05-25
adding alliance category mapping file to repo

sweng66:


sweng66:
This file will be used to map a pubmed type to Alliance category.

sweng66:
OK Thanks @azurebrd

PR #129, opened 2022-05-26
adding new field 'category' to json files

sweng66:


sweng66:
Will look into the functest tomorrow..

azurebrd:
@valearna @ianlongden The functional tests are failing because they're trying to read 
"FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/bin/src/literature/src/xml_processing/tests/pubmed_searches/pubMedType2allianceCategory_mapping.tsv'
And that's because the XML_PATH for tests has xml_processing/tests/ instead of just xml_processing/  and the pubmed_searches/ directory exists in xml_processing/ but not xml_processing/tests/   Is it better to copy pubmed_searches into tests, or would it work to symlink into tests, or would it be better to have the code check that XML_PATH ends in tests/ whenever it looks for files in 'pubmed_searches/' and strip 'tests/' before adding the rest of the path ?  I imagine both approaches work, but I don't know what's best practices.  Thanks !

azurebrd (approved):
That's great @sweng66 the logic makes sense.  As long as what Ian + Valerio say makes sense to fix the functional tests works, then it should be good :)

sweng66:
Cool @azurebrd 

will wait for Ian and Valerio to comment on it.

sweng66:
@valearna Yes, it is same for both dev and prod. I will try to fix the test(s) this PM if Ian has no concerns. 
Thanks Valeria!

azurebrd:
Thanks @valearna .  That's great @sweng66 .  I imagine you also want to do that in https://github.com/alliance-genome/agr_literature_service/blob/main/src/xml_processing/query_pubmed_mod_updates.py wherever you're creating a path that uses 'pubmed_searches/'   Even if it's not part of the tests (in theory it probably should get tested), it'd be correct in case the XML_PATH changes.

sweng66:
@azurebrd good point. Will add that when adding 'category' and 'language' to to the pubmed search program.

azurebrd (approved):
Cool cool @sweng66 .  Seems good to me.  Feel free to merge unless you want to wait for more comments.

PR #130, opened 2022-05-26
fixed to get rid of tests/ in the file path for false positive files …

sweng66:
…(just testing purpose

### DIFF
--- a/src/xml_processing/query_pubmed_mod_updates.py
+++ b/src/xml_processing/query_pubmed_mod_updates.py
@@ -227,7 +227,7 @@ def query_mods(input_mod, reldate):
         logger.info(f"Processing {mod}")
         fp_pmids = set()     # type: Set
         if mod in mod_false_positive_file:
-            infile = search_path + mod_false_positive_file[mod]
+            infile = search_path.replace("tests/", "") + mod_false_positive_file[mod]
             with open(infile, "r") as infile_fh:
                 for line in infile_fh:
                     pmid = line.rstrip()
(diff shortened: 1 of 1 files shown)


azurebrd:
@sweng66 Wouldn't it be better to change that in line 135-138 
search_path = base_path + 'pubmed_searches/'
search_outfile_path = base_path + 'pubmed_searches/search_new_mods/'
pmc_process_path = base_path + 'pubmed_searches/pmc_processing/'
pmc_storage_path = base_path + 'pubmed_searches/pmc_processing/pmc_xml/'

Maybe make a xml_processing path = base_path.replace("tests/", "")  and use that instead of base_path in those 4 lines ?
I can't tell what pmc_storage_path and search_outfile_path are for, but maybe they're used by other scripts downstream ?

sweng66:
@azurebrd Sounds good! Will do that. Initially, I thought this will only affect the input files.

azurebrd (changes requested):
Oh, no, I meant, make a new variable as a copy of XML_PATH without tests, and use that for the path variables that concatenate pubmed_searches/   Because base_path is used later on for json_filepath, json_path, log_path, and other things, and those should probably still be off of base_path instead of this tests-less path.

sweng66:
Is that just the original fix? just remove the 'tests/' from search_path if the 'tests' exists. @azurebrd

azurebrd:
Well, on lines 135 to 138 you're creating path variables using pubmed_searches/ based on the base_path which comes from environ XML_PATH and removing tests/ from all base_path,  but I think there's multiple cases where it should use the base_path from XML_PATH including the tests/   so that if you're running tests, you're generating files and loading files from the tests/ directory, not from the xml_processing/ directory.  That's why I thought you wanted a separate variable from base_path just for the paths that have pubmed_searches/ and strip out 'tests/' just for those.

azurebrd (approved):
Thanks @sweng66 !

PR #132, opened 2022-05-31
changed erratum to correction

sweng66:


azurebrd (approved):
Thanks @sweng66 !  I think it'll be best to confirm with Ceri and Kimberly that they don't want the obsolete category (I'm pretty sure we're not on the same page) to add that at the same time as `correction`.  So I'm holding off on that until then

PR #133, opened 2022-06-01
DO NOT MERGE -- SCRUM-1114 Actually SCRUM-606 but most comments are on 1114.  Adds re…

azurebrd:
…ferencecategory 'Correction' for pubmed type to alliance category mappings.  Adds volume mapping in dev_app

Do not merge.  Waiting for confirmation from Ceri and Kimberly that we do not also need an 'Obsolete' category.  

Making this PR for @ianlongden to look over. 

Also to ask @valearna and @FB-ping if it's okay to add the volume mapping to the dev-app to allow changing files inside-outside docker.  The dev server cannot run alembic outside docker, and alembic is needed to generate the version files, but I need to map them to exist outside docker to add to github.  If either of you think we should not permanently add the volume mapping to the dev-app, we don't have to, I can just do it locally when doing alembic things.

Also to ask @sweng66 to check that this makes sense for SCRUM-606  Adding corrections to reference_category schemas.


sweng66:
Adding 'Correction' to reference_category makes sense for pubmed updating script. @azurebrd

ianlongden (dismissed):

azurebrd:
Ceri and Kimberly agree we want to do it like planned before, deleting all the data from the obsolete paper, which will require an obsolete category, so I'll add it to this PR before merging. (do not merge yet)

sweng66 (approved):
Looks good! Thanks @azurebrd

ianlongden:
Not a problem for me, the final decision. It will be easier to code too :-)

PR #134, opened 2022-06-03
add pubmed update script and a functional test for testing first/corresponding author flag

sweng66:
When you are free, can you check out the code to see if there is any change I can make to improve the performance?
Also check to see if the functional test makes sense.. 

Thanks!


azurebrd (commented):
That's great @sweng66 !  It would have taken me soooo much longer to write that.  Learned a lot about sqlalchemy and queries too.

@valearna There's a couple of questions for you, but maybe it's best if Shuai thinks it over before we ask you, instead of you reading this review or looking at all that code.

Similarly there's a couple of questions for Ceri and Kimberly, that should probably be asked directly on the blue slack or a group message with the 4 of us.

I don't see a strikethrough option for stuff to cross out, I have some <strikethrough> for things that become irrelevant later (but left in to see thought process)

Review :  

I think Valerio thinks it's better to put the global variables between lines 22 and 35 inside main, or wherever is appropriate.  I forget why this mattered, but a while back there was something that wasn't working as global variables and we had to move them around in lot of scripts.

functional_tests.py update_data(None, pmid) not checking md5sum makes sense.  It general it should download all, generate md5sum, then only do updates if they're different.  I think the md5sum check is happening at line 302
              `if md5sum and pmid in old_md5sum and old_md5sum[pmid] == md5sum:`
is that right ?  I'm not sure if the functional test has old and new md5sums that are different, and the test would update it, but then sees that it has first/corresponding author and stops, or if the old and new md5sums are the same, so it doesn't try to update at all.  Maybe an additional test would be good to check that there's a PMID with an old md5sum different from a new md5sum (just make up a fake old md5sum ?) then update some database field with fake, try to update that PMID with the script, and see if the fake data is back to being correct ?

md5sum file should come from s3 using boto3, or is that getting downloaded by something else before the script runs ?

On sort_dqm_json_reference_updates.py there is
        env_state = environ.get('ENV_STATE', 'build')
        if env_state == 'prod':
            merge_md5dict = {}
            merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}
            save_s3_md5data(merge_md5dict, [mod])

I think
            merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}
merges any data in the old md5sum that isn't in the new one into the new one, and that's what a lot of `merge_md5sum` does ?  I think you want to upload the new md5sum file to s3 when done doing updates, with `save_s3_md5data`, but I realize we're just taking a look while it's in progress.  It should also only update the md5sum files when env_state is prod.


download_pubmed_xml already slices into 5000 papers at a time, just curious why have download_xml_max_size 150000 and slice that before passing to download_pubmed_xml .

strikethrough
I don't think I understand, why is it broken up by mod ?  I would have thought we'd want to update all of the pubmed papers in ABC together, not just
those for a specific mod.  I guess it's good to have the option to do only for one mod at a time, but then some papers that belong to multiple mods (about 10%, maybe) would get update multiple times ?  (This is all fine, just wondering)
/strikethrough (nevermind, see below)

I'm not sure how logging should work.  There's a logger where you're doing log.info, and that makes sense.  Separately you're writing to a log_file that ends in .log with fw.write ;  if you're pretty sure it's writing in the proper log format that should be what we need for logstash, then great, otherwise we should check with Valerio (or rewrite this if needed when working on the ticket about logs).

Great to see how to do complex queries through sqlalchemy =)

strikethrough
I'm guessing line 88 rename doesn't fail if the file doesn't exist ?  I tried it and got an error, but looking up about it, I only saw people asking about renaming when the destination file already exists, not when the source file doesn't exist.  I think at that point in the script, the md5sum doesn't exist, does it ?
/strikethrough Oh, you updated this while I was reviewing =P

Oh, I see the report is generated by MOD, so that when a curator from a mod needs to check something, it's only for their papers.  Got it !  Yeah, the potential overhead of re-processing some papers multiple times seems fine because of the generally low overlap.

I don't think you want to filter on ReferenceModel.mod_corpus_association because we want to update all the PubMed papers, even if they're not inside_corpus for any mod.  I guess it makes sense that curators would only want to update papers if they're in-corpus, but I would have thought we'd want to update all.  Maybe ask Ceri and Kimberly ?  Related to that, get_pmid_to_reference_id should check that  CrossReferenceModel.curie.like('PMID:%')  is not obsolete, otherwise we'll be updating data to obsolete PMID connections.  (So far this is the only thing I'm really sure should be different).

The only cross_references getting updated are DOI and PMC, and I think that makes sense.  We should probably confirm with Ceri and Kimberly that that's correct.  I can imagine curators manually changing those because PubMed is wrong, and then that getting overwritten, but it makes sense with the process that PubMed is the Source of Truth, even when they're wrong.

            # is it possible that this resource_id in database is a string?
I would have thought it was an integer because
   CREATE TABLE public.reference ( ...  resource_id integer, ...
 but I don't know.  Maybe type(x.resource_id) and type(new_resource_id) and see ?

I didn't realize ORCID were in cross_reference, although it makes sense since it needs to generate a URL from the curie.  I guess they're never connected to anything in the cross_reference table, they just have a corresponding string attached to the author ?

update_cross_reference_for_resource is written but not called by anything yet ?  That's okay.

I don't really know how sqlalchemy works, so I don't know db_session.add(x) works, but I assume you've queried x from the database (or created it from a model + data) and you're updating x, and doing .add(x) updates the database record by insert or update as appropriate ?

Thanks @sweng66 !

valearna:
I've taken a quick look at the code regarding (i) module-level variable declaration and (ii) logs. The issue we had with module-level variables and object initialization outside functions was that whenever we were importing functions from modules with such declarations we were also initializing objects that should have been initialized only when executing the module as top level code (__name__ == "__main__"). This means that it's ok to declare variables as module-level if they are not going to create issues or conflicts when the module or some of its functions are imported in other modules. For logs, I don't think we should focus on a single format as LogStash (or the exporters we'll use to send our logs there) probably supports multiple formats, even though using the standard logger with default formatting is probably the best solution at the moment.

sweng66:
Hi Juancarlos, thanks for looking over the code!! These are all good questions! I will try to explain what I think during my coding process. 

1. line 303 in update_pubmed_papers.py, this is happening only when mod is passed in (see line 300). 

In normal update case, if the old md5sum = new md5sum, it means no new update from pubmed. So we don't need to do anything (even though there is a first/corresponding author change for an author in the database).

In functional test, it won't check for old/new md5sum, so it will compare the data in the database and the data directly from pubmed to decide what to do next. Since in the functional test we have manually updated the data in the database before calling 'update_data' module so the program will decide if it needs to update the records. If there is a first/corresponding author flag, the script will only give a warning and will not update the authors for the given paper. So I think the functests should be able to catch this. Do you think so? 

2. About md5sum, I am not sure if the xml files in s3 match the data in our prod database. If yes, we can upload the files to s3 when the script runs against rdsprod (I can update code to do this).  

For this initial update, the script will download all xml files from pubmed and generate the json files for all of them, and compare the data from the database and the data from the json files to decide if there is a change (basically skip the md5sum checking since no previous versions of xml files/md5sum available). But after this initial run, all xml and json files and their md5sum will be in the system so later the monthly update will check the md5sum for each file in the system
(from the initial update) or in s3 (if we store them in s3) with the md5sum from newly downloaded/generated xml/json files and only proceed if there is a difference. 

3. About downloading xml files in 150000 batch,  when the script was downloading the xml files for MGI, it stopped in the middle (multiple times).. so I break it into 150000 xml files at a time and add a long break in between (10min break) to make PubMed "not so mad". =)

Not sure this is due to my slow internet connection. Maybe it is ok if it is running in cloud. 

4. About updating data by mod, just so we can have a clean report for each mod. Also there are so many papers so updating everything can take some time. Anyway, we can have a wrapper script to run it against each mod. Of course if we really want to run it against everything, we can add a 'no mod' option. I understand that some papers belong to multiple mods, but hopefully,  it shouldn't take too much extra time downloading them..

5. About logging, I agree. We should talk about this. The log_file that ends in .log with fw.write contains all the details about the update (old value, new value etc). The log.info part needs to be updated based on curators' specs.

6. About "I don't think you want to filter on ReferenceModel.mod_corpus_association because we want to update all the PubMed papers, even if they're not inside_corpus for any mod. I guess it makes sense that curators would only want to update papers if they're in-corpus, but I would have thought we'd want to update all. Maybe ask Ceri and Kimberly ? Related to that, get_pmid_to_reference_id should check that CrossReferenceModel.curie.like('PMID:%') is not obsolete, otherwise we'll be updating data to obsolete PMID connections. (So far this is the only thing I'm really sure should be different)."

I am not sure I understand this part, updating is based on mod so it will only retrieve the data that are needed and update the data associated with a given mod, right? Am I missing anything?   

By the way, good catch about the obsolete PMIDs!! I will add a check for this.

ps. how many papers are not associated with a mod ( if any) ?

7. In the initial update against local database, even though the resource_id looks the same but it still got updated (only happened for a few papers)  so I was confused.. anyway, not a big issue.

8. "update_cross_reference_for_resource is written but not called by anything yet ? That's okay" good catch!!  
   good catch! forgot to remove (wrote too many functions lol). will remove it.

9. "I don't really know how sqlalchemy works, so I don't know db_session.add(x) works, but I assume you've queried x from the database (or created it from a model + data) and you're updating x, and doing .add(x) updates the database record by insert or update as appropriate ?"

It works. you are right! x is a record "handler" from database through model so we can add/update data through it.  I code this way for sgd all the time.

sweng66:
Hi Valerio, Thanks for checking the code. Do you mean that the functional test is importing those python modules inside of  
"test_first_corresponding_author_flag' vs importing them in the beginning of the functional test script?  I was assuming these modules should not affect other tests.. ??

Yeah we need to come up with a std way to send reports. Right now, this script is calling the resource updating module, xml downloading, and json generating modules so we need to think about what to send to logger from each program to give curators a good summary of everything.

Thank you so much for spending time to review the code @azurebrd @valearna

azurebrd:
Hi @sweng66 ,

Thanks for breaking things down and numbering them :)

1 - Ah, okay.  So for now the old vs new md5sum check is only happen when given a list of PMIDs.  And when doing based on MOD, it's going to always update, and that's because it needs to update everything because of category and language, but in the future you will update it to also do that check ?
 Ah, got it, the functional test assumes that the data has changed, so it's not testing the old vs new d5sum.  That makes sense.

2 - The files in s3 should match the data in our prod database.  For some reason I can't access the s3 files with my account right now, so I can't doublecheck.  The original files should go into a directory that means original/ and the latest files should go into a directory that means latest/ .  We should also upload them as Glacier IR storage type, and I think the upload xml to s3 function already does this.  Sounds great about current run vs future runs.  Did what I write about `sort_dqm_json_reference_updates.py`  `merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}` vs `merge_md5sum` makes sense ?  It's okay if not, specially if you're not going to deal with it until the future update.

3 - Oh, good to know about the xml files failing.  I hadn't had that problem, but sounds good, as it's not really time-sensitive.

4 - Yeah, makes sense to keep things clean and simple to understand instead of adding extra code to check for which mod something belongs to and copying the output to several reports.  You're right that it shouldn't take long to re-process multi-mod files.

5 - Ah, gotcha.  I don't think the curators want the output of log.info because that has a bunch of timestamps and which script/function generated the line.  But they still haven't told us what they want, so who knows.  I thought the log.info would be what would get added to logstash, but I really don't know how that works (defer to Valerio, and we can do it on the other ticket)

6 - I thought we wanted to process all of our papers with non-obsolete PMIDs, instead of those belonging to mods.  When papers get created via PMID, they don't get a mod_corpus_association value (it's not a required field).  Same if a paper gets created manually and a PMID gets added to it later.  Possibly curators could assign papers to other MODs as needs_review and never check it, or even be out_of_corpus for all mods.  I suppose it'd be good
to check with Ceri and Kimberly whether they want all non-obsolete PMIDs updated, even if they're out_of_corpus or needs_review.  I agree that this makes the logs / reports much simpler.

About how many papers are not associated with a mod, some PMIDs got created through reference_comments_and_corrections and I think (but not sure) that they don't get attached to any mods, just to the paper that was its parent (or whatever connection).

I tried these queries on pgadmin, but it kept spinning and never returned anything.

SELECT * FROM public.reference WHERE reference_id NOT IN
( SELECT reference_id FROM public.mod_corpus_association )

SELECT * FROM public.reference WHERE reference_id IN
( SELECT reference_id FROM public.cross_reference WHERE curie ~ 'PMID' )
AND reference_id NOT IN
( SELECT reference_id FROM public.mod_corpus_association )

7 - yeah, weird.

8 - cool :)

9 - cool, thanks !

sweng66:
Hi @azurebrd 

1. Yes, when given a MOD, for current run, it will always update everything because of category and language like you said (also because there is no previous version of xml files available in the local system). For future runs (all category and language columns have been updated), the program will check new md5sum vs old one before updating (since all "old" version of xml files will be available in the local system in that time).

When given a list of PMIDs, the program will skip md5sum checking step and will always compare the data in the db vs data in json before updating. Assume the pmid list is not huge so comparing data in db vs in json won't cost much.

2. It is great to know that the data in db matches the data in s3. So instead of	checking old md5sum in the local system, the program will be updated to check the md5sum from s3 for the future runs. Thanks for the pointer, I will check out sort_dqm_json_reference_updates.py.

5. Agree.

6. The program is currently skipping the Obsolete papers and only updating the MOD associated ones, but I will think	about how to deal with the ones that are not associated with a MOD.

   I also tried those SQL queries and they never return anything.

azurebrd:
@sweng66 

1 - Ah, I'm not sure we can rely on the old xml and json files being in the system.  I originally wrote it similarly, but then the plan was to run it inside docker, and I wasn't sure the files would always be there.  At the time I didn't consider that we could have a docker volume from outside.  But that would mean it would always have to run from the same computer.  I'm not sure we need to compare the existence of files, since we can compare the md5sums, which will always be quickly downloadable from s3.  

Sounds good for PMID list always updating without md5sum check.

2 - Oh !  Great, thanks :)

4 - Cool ! 

I guess one could sqlalchemy the separate parts of the queries and do the joins in python.

azurebrd:
@sweng66 Got access to the account.  There's develop/ and prod/  , there's xml/ vs json_from_xml/ , and there's original/ vs latest/ 

Here's a URI where all the xml files should be.  Replacing the pairs of values above you can get the other versions.
s3://agr-literature/prod/reference/metadata/pubmed/xml/original/

I think the plan is to use develop just for testing a few at a time, so no need to upload to all the files every time a run happens.  But if you do, that's probably fine.  prod should always get all the files for live runs.

xml holds xml files and json_from_xml gets the files generated by xml_to_json from the xml files.

original gets the file the first time the PMID comes into the system.  latest gets the file every time a file gets generated after the first time.  The idea is that pubmed could delete the papers and then we lose the source of truth, so we want to keep a copy of  the most recent file, but also the original because sometimes authors won't get updated, so the latest file won't necessarily be the source of truth.

Thanks !

sweng66:
In RDSDEV:  

total pubmed papers:				                         689,012
Total pubmed papers associated with mod: 	                 683320
Total pubmed papers not associated with a mod:	    5,692

Will update these pubmed papers that are not associated with a mod.

azurebrd:
cool cool !  I guess we should check with Ceri and Kimberly about who should get notifications for those, but I imagine no one really wants them, and they don't matter too much.

sweng66:
yeah.

sweng66:
Just saw your comments about xml/json files in s3. Will look into it soon. Thanks @azurebrd

sweng66:
"merge_md5dict[mod] = {**old_md5dict[mod], **new_md5dict[mod]}
merges any data in the old md5sum that isn't in the new one into the new one, 
and that's what a lot of merge_md5sum does "

YES,  lol @azurebrd

sweng66:
@azurebrd Just looked at the code in sort_dqm_json_reference_updates.py for md5sum checking and saving..

Seems like the md5sum files are generated from dqm json files and they are MOD based,  one file for each mod.
Does that mean same paper from different mods may have different md5sum? 

The md5sum generated through 'generate_json' function will be different from the one from mod dqm file, I think, at least the allianceCategory and language fields are different. right?

Do you think we should upload a new md5sum file to s3 just for updating script?  This file will contain the md5sum for all pubmed papers in the prod database and it is generated through 'generate_json' function. We can start uploading this md5sum file along with updated xml files to s3 after we are done updating the rdsprod database for current run. 

And for all future updates, the updating script will download this md5sum file from s3 and compare with newly generated md5sum to check if a paper gets updated. In the end, the script will upload the newly generated md5sum file to s3 (only for prod update) and also upload the updated xml files to s3

The pubmed search program can  be fixed to add the md5sum for each new paper to this file as well.

Does this sound reasonable?  Or am I missing anything?

azurebrd:
Hi @sweng66 .  Yeah, there's multiple source of truth for each paper.  If the paper's at PubMed, most of the source of truth is the PubMed xml, but some stuff can only come from each MOD, like the mod_reference_type (I can't think of anything else right now, but I'm not looking at the code ;  for a while the allianceCategory was also coming from there), and for papers that are not at PubMed, then the mods are the whole source of truth, which yes, has conflicts when different mods have the same paper, but so far we only know of one paper created separately by two mods, and we only know they're the same paper because they have the same DOI (since many papers can have the same title, or be the same paper but have the authors in a different format, and so on, so it's really hard to know whether any given two papers in our database are actually different papers).  

For each paper that comes from each source (a mod or pubmed), there's an md5sum file https://s3.console.aws.amazon.com/s3/buckets/agr-literature?prefix=develop%2Freference%2Fmetadata%2Fmd5sum%2F&region=us-east-1&showversions=false# (and also a prod version for the prod files, which is what should really be used).  When doing any updates from any source, the idea is to download its md5sum file, then for each new paper generate its new md5sum, and only if it's different from the s3 file md5sum, do we update the data in ABC, then update the md5sum file and upload it back to the s3 location.  I think this is basically what you said, but when you said "Do you think we should upload a new md5sum file to s3 just for updating script?" I wasn't sure if you meant that we needed to create a whole new type of file for this purpose.  I think we want to use the existing file, then update it, and then yes, upload this new file to s3 for future updates.  Please let me know if this isn't clear.  Thanks !

sweng66:
@azurebrd Oh cool. I didn't know the file is already in s3.  I don't have access to the alliance AWS account so I am not sure what is in the s3. Sorry for the confusion. I will check things out and fix the program to use this file. Thanks!!

About AWS access,  who should I talk to? Adam Wright?

sweng66:
@azurebrd Adam just setup the passwd for me to access aws from console so I am able to see what is in there now =)

https://s3.console.aws.amazon.com/s3/buckets/agr-literature?prefix=develop%2Freference%2Fmetadata%2Fmd5sum%2F&region=us-east-1&showversions=false#

I saw the file PMID_md5sum

Assume this is the file the updating script should use. Correct?

sweng66:
@azurebrd Just checked xml_to_json.py. It is awesome to know that this program is updating the PMID_md5sum every time it is generating the json files so  the PMID_md5sum in s3 is always up to date as long as the PubMed papers are loaded/updated through this pipeline.

azurebrd:
@sweng66 Oh, cool that Adam got you set up !  In theory Valerio or I could also set you up, but I don't know how, so we would have had to ask Valerio.

Yeah, that would be the file to update when doing testing in develop against rdsdev.  And then you'd change the 'develop' to 'prod' for live changes.

Oh !  I'd forgotten that xml_to_json does that update !  That's great.  It should always be correct for that to update the md5sum file.  The uploading of the xml and json files should still happen separately to the correct s3 locations based on develop/prod original/latest (first time vs update) and xml/json .  Please let me know if that doesn't make sense.  I don't think I wrote code to already handle all that (certainly not for updates, since that wasn't happening yet, but there's probably code for develop/prod and xml/json)

sweng66:
@azurebrd Yes, we still need to upload the xml and json files and I will let you know if I need any help. Thanks

PR #135, opened 2022-06-06
Scrum 1571 - debezium sync on reference table

valearna:
This PR adds debezium to sync elasticsearch with postgres. For now only the reference table is synced (no joins with related tables such as xref or author). Debezium is added through a set of docker-compose services - kafka, zookeeper and debezium connector are standard images, whereas debezium setup is used to send the configuration needed to set up the elasticsearch index (settings and mappers) and the debezium connectors (postgres for data in and es for data out). New entries in the Makefile simplify the deployment.


azurebrd (approved):
Thanks @valearna !

sweng66 (approved):
Looks good!

PR #136, opened 2022-06-07
Scrum 1571 - fixed makefile

valearna:
.env file can now be passed as a parameter to debezium makefile scripts


azurebrd (approved):
Thanks !

PR #138, opened 2022-06-08
fixed to not update 'category' and 'language' columns if it is a pubm…

sweng66:
…ed paper

### DIFF
--- a/src/xml_processing/sort_dqm_json_reference_updates.py
+++ b/src/xml_processing/sort_dqm_json_reference_updates.py
@@ -623,6 +623,7 @@ def update_db_entries(headers, entries, live_changes, report_fh, processing_flag
     # fields_simple_camel = ['title', 'allianceCategory', 'citation', 'volume', 'pages', 'language', 'abstract', 'publisher', 'issueName', 'issueDate', 'datePublished', 'dateLastModified']
     # removed some fields that Ceri and Kimberly don't want to update anymore  2022 04 25
     fields_simple_camel = ['title', 'allianceCategory', 'volume', 'pageRange', 'language', 'abstract', 'publisher', 'issueName', 'datePublished']
+
     # there's no API to update tags
 
     api_port = environ.get('API_PORT')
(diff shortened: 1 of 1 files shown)


azurebrd (commented):
Hi @sweng66 .  This is part of update_db_entries where processing_flag == 'mod_biblio_all', which can only happen if line 422 flag_aggregate_biblio and not flag_aggregate_mod_specific.  

Starting at line 375, it loops through all the cross references, to see if any of them are a PMID, because while the primaryId _should_ be the PMID if there is one, it's not required, and I don't think the DQM validator forces that, so the DQMs could send a PMID in the cross_reference and have primaryId be whatever.  Line 391 sets flag_aggregate_mod_specific true if any have a PMID, and line 395 sets flag_aggregate_biblio true if any have a MOD id.  

Since your changes only happen when mod_biblio_all + flag_aggregate_biblio (has_mod_id) + !flag_aggregate_mod_specific (!has_pmid), the prefix can never be PMID in line 699, so I think that check is unnecessary, and potentially confusing because it could seem that it's checking something that couldn't happen.  It might be a good check to have if the previous processing I described changes to no longer work that way.  But then it assumes the PMID would be the primaryId, which is not a safe assumption (which is why that processing in 2nd paragraph happens).  

The 2nd paragraph processing isn't obvious, so if what I said makes sense, it might be helpful to add comments to the code to explain what it means, so that future developers can better understand it.  Sorry I didn't already do that, but I wasn't sure how much to say or where to say it.  

Either way, I'm sure the code in the PR works, so if you want it this way, I can approve it.  If what I said gave you some context and you can think of a better way to write it, then good.  But again, it's totally fine as-is, I just wasn't sure the 2nd paragraph processing was obvious.  Thanks Shuai =)

sweng66:
It is confusing. Lol 

Can I reassign it to you?

sweng66:
I thought primaryId will always be PMID if it is a pubmed paper

azurebrd:
Haha, I think it's already okay as-is (without making any changes to the code, but I didn't know that until I was reviewing what you did, sorry about that).  

The dqm_entry['primaryId'] is what comes from the DQM files directly, right, it hasn't been processed and modified by the code ?  The DQM _should_ use the PMID as the primaryId if the paper has a primaryId, but in the past I've uploaded files that weren't doing that, so I know it's possible, and I haven't tried to validate that if there's a PMID in cross_references, that it is also the primaryId.  There might be something like that in parse_dqm_json_reference.py but I'm not sure.  Since I was trying to account for DQMs potentially not putting the PMID in the primaryId, I wrote the stuff in the previous paragraph 2, but I wasn't sure how to document it.  I think currently the script already doesn't get to line 699 if there's a PMID, so I think the check isn't adding anything, but it might still be useful to have later, so I can approve it.

azurebrd (approved):
Just approving

sweng66:
Yes, I just checked the code. The addition doesn't do anything. I will close this PR. Thanks for clarification.

azurebrd:
@sweng66 Oh, not at all.  Sorry I didn't realize that before you wrote it.  It would probably still be helpful to document whatever you figured out.  I liked the comment you had in the code on this PR, it seemed helpful.  The whole xml_processing would probably use more comments or documentation.

PR #139, opened 2022-06-08
Scrum 1571 - added elasticsearch index name as a parameter to the API

valearna:
This PR adds the env var ELASTICSEARCH_INDEX to the API to select the es index to connect to. The var is now required by the API and should be added to all .env files in use.


azurebrd (approved):
Thanks @valearna !  Let me know if I should pull and restart the APIs that we have running for rdsdev  and rdsprod

valearna:
thanks @azurebrd! No, no need to restart the API, I restarted it with code from this branch before opening the PR to test it and it's still running

azurebrd:
Cool, thanks @valearna !

PR #140, opened 2022-06-13
Scrum 1477 - New project structure

valearna:
- Moved both api and lit_processing (fka xml_processing) under the same python dir so that they can be imported using a single PYTHONPATH
- Fixed all mentions to old locations in scripts
- Alembic moved outside api
- requirements files moved to project root
- Fixed additional mypy errors


azurebrd (approved):
Thanks @valearna !

PR #142, opened 2022-06-21
adding report to the pubmed update script and make it cron ready 

sweng66:


azurebrd (commented):
Thanks @sweng66 , that's great !
I see there's a few new variables coming from environ https://github.com/alliance-genome/agr_literature_service/blob/bfb9dea779d6f4c86717c76a5dfa6c022399f0d5/agr_literature_service/lit_processing/update_pubmed_papers.py#L870 but I don't see changes to .env   Do you think they need to go there, or should there be a different way of getting them in ?  @valearna might also have an opinion on what should go in the .env and what shouldn't.

sweng66:
@azurebrd Thanks for checking out the code. =) 
Good catch! I will add these these variables to .env (just a place holder for each since I don't think we want to expose these to the public)

sweng66:
Just added these new variables into .env file. Let me know if you have any concerns. Thanks

sweng66:
Or should we keep these variables in our own local .env* just like XML_PATH etc?

azurebrd:
I don't think we want `export` in the .env file.  I'm also not sure about having the email parameters in there, because that would more or less lock us into using 1 set of emails for every script that sends emails, right ?  Or require a lot of different variables in the .env file for different scripts having different emails ?

sweng66:
Good catch!  @azurebrd  I was copying / pasting lol.  Anyway got rid of "export" from .env file. I assume we will send all reports from cron jobs to lit-cron-jobs channel (CRONTAB_EMAIL), at least for literature related data loading / updating. Right? Am I missing anything?  I just don't want to expose the EMAIL address in the code to public and can't think of a better place to store this variable.  Any suggestions?

azurebrd:
cool cool.  Yeah, it makes sense that all the emails to slack would go there.  I guess eventually if curators want direct personal emails, we'd have to figure something out for that ?  There's a way to store values in secret like the AWS keys, but I don't recall how that works.  We could also put them in the database, but that seems like a lot of overheard.  Maybe we'd do that when we have Person objects and only email to personal curator email addresses at that point ?

azurebrd:
Also, I'm not sure about XML_PATH   It used to be one of the .env variables, but with @valearna we found that in docker-compose it was always hardcoded, so I don't know.  Might be best to talk about it together, the three of us ?

sweng66:
@azurebrd yes, we should talk about it. In sgd, we set all private variables in ENV and for docker containers, we store these env files in private s3 bucket.

PR #145, opened 2022-06-27
Scrum 1572 Join reference and cross_reference in kafka to create combined elastic search index

valearna:
This PR adds a new Kafka connector to read the cross_reference table and uses ksqldb to create ktables from Kafka streams, collapse xrefs into a single list for each reference_id and join references and lists of xrefs into a single ktable, which is then sent to a new 'reference_joined' Kafka topic, feeding it to elasticsearch.

Tested editing reference title and deleting a reference - changes are reflected in elasticsearch


debezium/ksql_queries.ksql
```diff
@@ -0,0 +1,84 @@
...
+    date_published string,
+    date_updated bigint,
+    issue_name string,
+    keywords array<string>,
+    language string,
+    open_access boolean,
+    page_range string,
+    plain_language_abstract string,
+    publisher string,
+    pubmed_abstract_languages string,
```
azurebrd, line 23: This should probably be array<string>  
in sql     pubmed_abstract_languages character varying[] COLLATE pg_catalog."default",

azurebrd (approved):
Thanks @valearna !

PR #146, opened 2022-07-02
Scrum 1695

ianlongden:


ianlongden:
It may need some tidying but the test work and it should be enough for UI code to come.

agr_literature_service/api/tests/test_017_reference_tag.py
```diff
@@ -0,0 +1,105 @@
...
+from sqlalchemy.orm import sessionmaker
+
+from agr_literature_service.api.crud.reference_tag_crud import patch
+from agr_literature_service.api.crud.mod_crud import create as mod_create
+from agr_literature_service.api.crud.reference_crud import show, create as ref_create, patch as ref_patch
+from agr_literature_service.api.database.config import SQLALCHEMY_DATABASE_URL
+from agr_literature_service.api.database.base import Base
+from agr_literature_service.api.models import ModModel, ReferenceModel
+from agr_literature_service.api.models.reference_tag_model import ReferenceTagModel
+from agr_literature_service.api.schemas import ReferenceSchemaPost
```
azurebrd, line 11: Is this redundant with line 12 ?

agr_literature_service/api/tests/test_017_reference_tag.py
```diff
@@ -0,0 +1,105 @@
...
+db.execute('delete from cross_reference')
+db.execute('delete from author')
+db.execute('delete from editor')
+db.execute('delete from "reference"')
+db.execute('delete from resource')
+
+ref_curie = "AGR:AGR-Reference-0000000001"
+
+
+def get_ids():
```
azurebrd, line 38: Not in use ? (commented out code later)  That's ok

azurebrd (commented):
Thanks @ianlongden .  If I'm understanding this right, one can only create tags through creating or patching the reference, as opposed to the other tables that are separate like mod_corpus_association, having direct _crud ways to create / destroy / patch ?

I can't tell how we would remove a tag that was entered by mistake, the reference_crud passes each tag to patch_reference_tag to update, but it doesn't delete the tags that weren't sent in (because it's a patch, not a put), and there's no explicit way to delete a specific tag by either its id or it's mod-type pair ?  If it does delete all tags that aren't in the patch request, that might be bad, because different UIs might only be handling mod-specific or type-specific tags, but I suppose they could always keep track of all tags in the reference and send in the non-modified tags as well.

Is it better to do things this way (only through a reference patch) ?  Because the other tables have separate endpoints, which the UI is working with, so it seems non-standard to do things differently for different tables.

Why did we lose test_016_mod_corpus_association ?

azurebrd:
@ianlongden says it's ok to close this

PR #147, opened 2022-07-05
API for downloading json files from s3

sweng66:


docker-compose.yaml
```diff
@@ -144,8 +144,6 @@ services:
       - agr-literature
     volumes:
       - agr-logs:/logs
-    depends_on:
```
azurebrd, line 147: @sweng66 I think you don't want to remove that from the api docker-compose for main, but you did need that to run that on the devserver

azurebrd (commented):
That's great @sweng66 !  
I think other than that postgres in docker compose getting removed, it's all good.
What are aiofiles and starlette ?  I don't see them in the changed files, so I'm wondering whether you tried them out and didn't use them, or whether they're used but don't need to be imported ?  Thanks !

docker-compose.yaml
sweng66: @azurebrd  got it. I changed it back! thanks

sweng66:
@azurebrd good catch! the fastapi needs aiofiles and starlette to be able to work with FileResponse properly.

valearna (approved):
Everything looks good, thanks @sweng66! I would pass 'bucketname' as an env variable and set it in the .env files but it's up to you if you want to do this now or later.

sweng66:
@valearna Good suggestion!! I just added 'BUCKET_NAME' to .env and fixed the download.py to get bucketname from ENV.

azurebrd:
Cool, thanks for explaining @sweng66 !   @valearna Passing the bucketname sounds fine to me, but it's always going to be agr-literature for the literature service, right ?  I guess this allows for in the future having a different root-level bucket that isn't for literature where the literature service can deposit and get files from ?

valearna:
The bucket name could be defined in the code and doesn't need to be in the .env file if it's always going to be the same, but we could define it once (maybe in the api/config/GlobalConfig class?) and import it from the scripts instead of defining it inside each scripts.

azurebrd:
That makes sense

sweng66:
Thanks both for the explanation! @valearna @azurebrd   

Everything makes sense to me. The current fix is to use "agr-literature" as default bucket name if nothing is passed in from ENV for this variable so I will leave as is. Later we can 'globally' fix every program/module (that is using s3) to get the bucket name from a config file at the same time. Does this sound reasonable?

azurebrd:
Cool, sounds good @sweng66 !  Let me try to revert main before merging this, so we probably shouldn't update this branch with main.

azurebrd (dismissed):
Thanks @sweng66 and @valearna !

.env
```diff
@@ -52,6 +52,8 @@ DEBEZIUM_KSQLDB_PORT=8088
 OKTA_CLIENT_ID=
 OKTA_CLIENT_SECRET=
 
+BUCKET_NAME=
```
valearna, line 55: We should remove this variable from here

valearna (approved):
Everything looks good now, thanks @sweng66!

PR #148, opened 2022-07-07
Scrum 1568

azurebrd:
Mostly adding xenbase and handling their crossReference types.  But also in parse_dqm_json_reference.py moving the logic for tracking dqm_xrefs prefixes and identifier for all xrefs to just valid xref prefixes.  Also account for dateArrivedInPubmed with different casing, and exclude that when coming from DQMs because it should only be populated from PubMed.


sweng66 (approved):
Looks good! Thanks Juancarlos ! @azurebrd

azurebrd:
Thanks @sweng66 !

PR #149, opened 2022-07-07
Scrum 1573 - pubmed query automation with docker + cron

valearna:
This PR adds a new docker-compose service that runs cron and automates scripts


sweng66 (approved):
Looks good!! thanks Valerio.

azurebrd (commented):
Was the tree requirement because looking at a directory didn't work in dev-shuai as opposed to working now with a symlink in public_html ?   If you both like tree better, cool, but it might not be needed anymore (totally fine to leave it, if it's a hassle to remove it and test.  It does look nicer than the default browser view of a directory.

PR #151, opened 2022-07-08
Remove 'tree' command from the PubMed update and search programs

sweng66:


valearna (dismissed):
The code is much better without tree, thanks @sweng66

valearna (approved):

sweng66:
The fix for both scripts are working fine so merging it now

azurebrd (commented):
Thanks Shuai and Valerio

PR #153, opened 2022-07-11
Scrum 1615 Debezium on Author table

valearna:
- Loaded author table in kafka and ksqldb through Debezium
- created materialized view reference_xref
- joined author table with reference_xref


valearna:
Thanks @azurebrd and @sweng66 I removed the volumes because I was both loading the files in the docker image and passing them as volumes

PR #154, opened 2022-07-12
feat(debezium): join mod_corpus_association table

valearna:
- created two additional fields in the final ES objects:
  1. mods_in_corpus: the list of mod abbreviations for which the ref is in corpus
  2. mods_needs_review: the list of mod abbreviations for which the ref is in needs review
- now using left outer joins on all joined tables to support refs with null values (no authors, no xrefs or no mod_corpus_associations)

### DIFF
--- a/debezium/ksql_queries.ksql
+++ b/debezium/ksql_queries.ksql
@@ -51,6 +51,26 @@ CREATE TABLE author (
     VALUE_FORMAT='json'
   );
 
+CREATE TABLE mod (
+    mod_id string PRIMARY KEY,
+    abbreviation string
+  ) WITH (
+    KAFKA_TOPIC='abc.public.mod',
+    VALUE_FORMAT='json',
+    KEY_FORMAT='json'
+  );
+
+CREATE TABLE mod_corpus_association (
+    mod_corpus_association_id string PRIMARY KEY,
+    mod_id string,
+    reference_id string,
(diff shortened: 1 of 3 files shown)


azurebrd (approved):
Thanks @valearna !   The tables seem to keep getting longer and longer, as expected.  I wonder if it'd make more sense to start with the smaller table and join the reference table at the end since it has the most fields.  Not in a logical makes-more-sense, but in save-space makes-more-sense.  Or maybe if the separate tables should get a numeric code or greek letter code, to avoid very long table-join-composite names (and end up with 001_002_003 for the join table, where 001 is 001_reference, 002 is 002_authors).  Just some thoughts, not saying it'd be better.  I personally like long descriptive variable names.

valearna:
Thanks @azurebrd, unfortunately the reference table needs to be joined first, as for the current limitations in ksqldb table-table joins can only be performed on a field which is primary key (and ROWKEY) for the left table! I tried to select * instead of specifying the full list of fields every time but that didn't work as ksqldb was renaming them by attaching the table name to the field name. Hopefully the list of tables to join won't grow much longer than this...

valearna:
oh, and I prefer to keep descriptive names, they are long but at least it's easy to understand what each table contains

azurebrd:
Cool, thanks @valearna !

PR #155, opened 2022-07-13
SCRUM-1691 API changes return obsolete_reference_curie data in the re…

azurebrd:
…ference show crud as "obsolete_reference" array of curie strings.


sweng66 (approved):
Looks great! Thanks @azurebrd

PR #157, opened 2022-07-14
SCRUM-1691 Forgot to change schema for the API, maybe cleaner to remo…

azurebrd:
…ve obsolete_reference since adding obsolete_references


sweng66 (approved):
Looks good!

PR #159, opened 2022-07-15
Scrum 1357 New facets for "in corpus" and "needs review"

valearna:
- Facets added to search api endpoint
- always show all facet values even when count = 0


sweng66 (approved):
Looks good!

valearna:
thanks @sweng66 !

PR #160, opened 2022-07-15
Refactor dump script to improve performance and CPU/memory usage

sweng66:


agr_literature_service/lit_processing/dump_json_data.py
```diff
@@ -526,24 +498,24 @@ def get_journal_data(db_session):
 
 def print_usage():
```
valearna, line 499: Argparse has its own print_help() function and we can avoid defining an additional one. If the only reason to add this function is to force `mod` to be in a predefined list of allowed values, we can use the `choices` option for add_argument(): https://docs.python.org/3/library/argparse.html#the-add-argument-method

valearna (commented):
The new way to read data from the DB with raw sql queries instead of the ORM looks good

valearna (approved):

sweng66:
Thanks Valerio!

PR #162, opened 2022-07-16
adding XB to the update list, clean up argparse

sweng66:


valearna (commented):
The PR Looks good. I wonder whether we should define the list of mods in a single location or read them from the DB. Any thoughts @azurebrd @ianlongden ?

valearna (approved):

azurebrd (approved):
@valearna Yeah, we do have a hardcoded list of mods in a bunch of scripts and a bunch of places in the UI.  In the UI sometimes they need to be the abbreviation vs the curie prefix, so we'd need to extend the mod table to have a curie_prefix column first.

Edit: or at least define the list of mods in a separate file in one place and import that.

sweng66:
We have 'GO' in the mod table. Are we going to store the GO related data in the database? 

we can fix the data dumping and pubmed update scripts to get a list of mod abbreviations (excluding 'GO') from  the database.. but still need to hardcode 'GO' in the code unless we are adding a new column to mark that in the mod table.

sweng66:
@azurebrd Thanks for checking out the fix!!

azurebrd:
Oh, yeah, good point, GO is not meaningful for the lit_processing, but is still a mod.

PR #165, opened 2022-07-19
cleaned up the sql queries and added an email sending section for demanded file download

sweng66:


azurebrd (dismissed):
Oooh, nice, having an array with the column headers to join together.

It's kind of odd that in send_email_report it gets status, meaning what happened with generating the file, then sends an email which also has a status, reusing the status variable, instead of a different variable name, like status_email vs status_file_generation (or something).  But it all looks good.

Thanks @sweng66 !

sweng66:
@azurebrd Good catch! omg I didn't realize that I used the same exact variables in same function.  Fixed.

azurebrd (approved):

PR #166, opened 2022-07-19
adding an end point for downloading the reference file by filename

sweng66:


azurebrd (approved):
Looks good, @sweng66 !

PR #173, opened 2022-07-20
feat(docker-compose): removed dependency on postgres from api

valearna:


azurebrd (dismissed):
Thanks @valearna .  Can we also remove it from the dev_app and still have the dev_app work ?

valearna:
Removed postgres dependency from dev-app and added command to start postgres in the Makefile for dev-app commands

azurebrd (approved):
That makes sense, thanks @valearna !

PR #174, opened 2022-07-21
SCRUM-1691 API fix for patch of reference_comment_and_correction_crud…

azurebrd:
…, reference_comment_and_correction_update object should not have dict() applied to it to get its items.


valearna (approved):
Thanks @azurebrd, is this related to differences in the specific python version we are using in docker vs. the one used for development?

azurebrd:
Thanks @valearna .  I don't think so, I think that endpoint was just broken.

PR #176, opened 2022-07-21
Scrum 1438 final

ianlongden:


agr_literature_service/api/crud/reference_crud.py
```diff
@@ -409,3 +416,92 @@ def merge_references(db: Session,
...
+               journal, volume, issue, page_range)
+    return citation
+
+
+def author_order_sort(author: AuthorModel):
+    return author.order
+
+
+def citation_from_data(reference_data, authorNames):
+    authorNames = authorNames[:-2]  # remove last '; '
```
valearna, line 435: Can we remove the last '; ' in the main crud function or leave it here but check if `authorNames.endswith('; ')`?

valearna (approved):
This looks great, thanks @ianlongden

agr_literature_service/api/crud/reference_crud.py
```diff
@@ -108,6 +110,8 @@ def create(db: Session, reference: ReferenceSchemaPost):  # noqa
                         obj_data["orcid_cross_reference"] = cross_reference_obj
                     del obj_data["orcid"]
                     db_obj = create_obj(db, AuthorModel, obj_data, non_fatal=True)
+                    if db_obj.name:
+                        authorNames += db_obj.name + '; '
```
azurebrd, line 114: Are these authors at this point ordered by the 'order' field ?   I see that happens in get_citation_from_obj but not in get_citation_from_args, which uses authorNames from here ?

azurebrd: Oh, this is only when creating a reference, so the authors are already ordered.

agr_literature_service/api/crud/reference_crud.py
```diff
@@ -409,3 +416,92 @@ def merge_references(db: Session,
...
+def get_citation_from_args(authorNames: str, year: str, title: str, journal: str,
+                           volume: str, issue: str, page_range: str):
+    # Create the citation from the args given.
+    citation = "{}, ({}) {} {} {} ({}): {}".\
+        format(authorNames, year, title,
+               journal, volume, issue, page_range)
+    return citation
+
+
+def author_order_sort(author: AuthorModel):
```
azurebrd, line 430: This isn't being used ?

alembic/scripts/add_citation.py
```diff
@@ -0,0 +1,69 @@
...
+    print("Starting {}".format(datetime.datetime.now()))
+    outfile = open("citation_update.txt", 'a')
+    seen = {}
+    i = 0
+    last_seen = 0
+    batch_size = 1000
+
+    descending = db.query(ReferenceModel).order_by(ReferenceModel.reference_id.desc())
+    last_ref = descending.first()
+    print("Last referecne_id is : {}".format(last_ref.reference_id))
```
azurebrd, line 42: referecne_id -> reference_id 
not important

azurebrd (approved):
Thanks Ian

alembic/scripts/add_citation.py
ianlongden: I will fix typos etc later.

agr_literature_service/api/crud/reference_crud.py
ianlongden: I will revisit when I do more work. 
But yes that makes sense.

agr_literature_service/api/crud/reference_crud.py
ianlongden: Yep, I changed the way I did things in the end. Will Remove later.

PR #178, opened 2022-07-23
fixed a bug in comments/correction section

sweng66:


azurebrd (approved):
Looks good @sweng66 !

PR #181, opened 2022-07-25
adding code to update citation

sweng66:


valearna (approved):
Everything looks good to me, thanks @sweng66

ianlongden (commented):
That's a lot of code when I could have simply added return True/False.

sweng66:
@valearna @ianlongden Thanks for looking over the code!
@ianlongden The script already has the reference/author/resource data in the memory for generating the citation. While if the script is calling your citation update module, it will need to go to database to retrieve these data again for each updated paper.  It will probably slow down the PubMed updating program. I see your point that we have some repeating code here (eg format year and authors). Maybe I should fix the script to just directly call your citation update function. 

What do you all think?

ianlongden:
Okay if there is a good reason then I am happy :-)

azurebrd:
@ianlongden .  Valerio and Shuai were talking, and the plan is to rewrite citation_from_data to take in the author array and the date, then do the parsing into a string for all authors and the year, then the lit processing code can use the same function.  (Might have missed some details, but that's the general idea)

ianlongden:
Makes sense. Thanks,

sweng66:
@ianlongden I just fixed the 'get_citation_from_args' function in reference_crud.py so the command line scripts can also use it without repeating code. Can you check to make sure it won't mess up anything? I have tested and it worked fine for pubmed updating program. Thanks!

azurebrd (approved):
Thanks, @sweng66 !

ianlongden:
That’s what the tests are for :-)
But thanks for checking with me.

-Ian.

PR #182, opened 2022-07-26
Scrum 1810 - add citation field to elasticsearch through debezium and kafka

valearna:
Added new citation field to kafka tables and joins


sweng66 (approved):
Looks good!

PR #183, opened 2022-07-26
Scrum 1583

ianlongden:


agr_literature_service/api/crud/reference_crud.py
```diff
@@ -288,6 +300,7 @@ def show(db: Session, curie: str, http_request=True):  # noqa
         else:
             return None
 
+    print("REF: {}".format(reference))
```
azurebrd, line 303: superfluous ?

agr_literature_service/api/crud/reference_crud.py
```diff
@@ -304,6 +317,8 @@ def show(db: Session, curie: str, http_request=True):  # noqa
         reference_data["cross_references"] = cross_references
         # del reference_data["cross_reference"]
 
+    # So this wierd we check reference.mod_reference_type BUT
```
azurebrd, line 320: where do you see this ?  It looked that way in the PR when some lines were collapsed, but expanding it, new lines 322-328 seems to be about MRT and 334-342 about MCA

agr_literature_service/api/crud/reference_ontology_crud.py
```diff
@@ -0,0 +1,197 @@
...
+        reference_ontology_data["reference_curie"] = db.query(ReferenceModel).filter(ReferenceModel.reference_id == reference_ontology_data["reference_id"]).first().curie
+    del reference_ontology_data["reference_id"]
+    if reference_ontology_data["mod_id"]:
+        reference_ontology_data["mod_abbreviation"] = db.query(ModModel).filter(ModModel.mod_id == reference_ontology_data["mod_id"]).first().abbreviation
+    del reference_ontology_data["mod_id"]
+
+    return reference_ontology_data
+
+
+def show_by_reference_mod_abbreviation(db: Session, reference_curie: str, mod_abbreviation: str) -> list:
```
azurebrd, line 145: This seems pretty useful.  Is it used somewhere ?

agr_literature_service/api/tests/test_001_reference.py
```diff
@@ -180,14 +183,42 @@ def test_reference_large():
                 ]
             }
         ],
+        "ontologies": [
+            {
+                "ontology_id": "Ontology1",
+                "mod_abbreviation": "RGD_ont",
```
azurebrd, line 189: Oh, does posting this full_xml create a MOD entry with that mod_abbreviation ?  I would have thought it would fail because it doesn't exist.  Or do the RGD_ont + FB_ont mods get created sometime before the first test runs ?

agr_literature_service/api/tests/test_017_reference_ontology.py
```diff
@@ -0,0 +1,163 @@
...
+def test_get_bad_ref_ont():
+    with pytest.raises(HTTPException):
+        show(db, 99999)
+
+
+def test_bad_missing_args():
+    xml = {'reference_curie': "AGR:AGR-Reference-0000000001",
+           'ontology_id': "ont1",
+           'created_by': "017_Bob"}
+    with pytest.raises(ValidationError):  # mod_abbr missing
```
azurebrd, line 63: I'm pretty sure they want to be able to not put in a MOD, and have null mean all mods.  Does this mean that ReferenceOntologySchemaCreate having      mod_abbreviation: str  instead of      mod_abbreviation: Optional[str], require that create a referenceOntology entry must pass a mod_abbreviation: null, or that it cannot allow a mod_abbreviation: null ?  (If it requires passing an explicit null, that's fine, if not, it's probably not right)

azurebrd (approved):
Looks good, @ianlongden !  I had a few questions as single comments.  Glad to see how you did all the testing.  Thanks.

agr_literature_service/api/tests/test_017_reference_ontology.py
ianlongden: Good spot, I need to test this and see what happens.

agr_literature_service/api/crud/reference_crud.py
ianlongden: Yep commented at the wrong line I will correct comment and where it is too:-
    # So thisis wierd, we check reference.mod_corpus_association BUT
    # use reference_data["mod_corpus_association"]
    if reference.mod_corpus_association:
        for i in range(len(reference_data["mod_corpus_association"])):

azurebrd:
Also, not sure when we'll deploy the citation alembic changes, but we might consider not merging this into main until that one has alembic applied to prod.  Might not matter though if both are ready for prod at once, or a different commit can be used to apply alembic to prod, or alebic can upgrade to a commit that's not head.  But would be easier if it was all in main ?  (maybe not)

ianlongden:
Ceri has requested some changes so do NOT merge for now. Plus I would like to do the tests as mentioned earlier, just to make sure...
I presume you can upgrade to a specific point by doing alembic upgrade XXXXX rather alembic upgrade head.

agr_literature_service/api/tests/test_017_reference_ontology.py
ianlongden: Tested, you need to pass "mod_abbreviation": "", if you to not specify the mod (i.e. all mods)

azurebrd (approved):
Thanks @ianlongden !   Let's hold off on merging to main until we discuss deploying the citation change to prod at planning though.

PR #185, opened 2022-07-27
adding code to update citations

sweng66:


azurebrd (approved):
Thanks @sweng66 !   So if the md5sum is different and it maps to exactly one AGR curie, it updates it, even if no data associated with the citation is what changed.  This seems fine to me.  Did the pubmed update script generate the citation and compare it before triggering the update ?  I think it'd be fine both way, so this is good.  Thanks !

sweng66:
@azurebrd Thanks for checking out the code. 

Yes, the update_citation function will compare old/new citations before updating. Since loading dqm data is pretty fast, updating citation through update_citation function will be fine. Too lazy to get out all related fields for every updated paper.

sweng66:
@azurebrd The pubmed update program will create a citation for each updated paper and compare the newly generated citation with the one in the database and update it if there is a change. Since we need to output the citation change count in the report and also for performance consideration.

azurebrd:
Oh, got it, thanks @sweng66 !

PR #186, opened 2022-07-28
SCRUM-1694 add router

ianlongden:


azurebrd (approved):
Thanks @ianlongden

ianlongden:
Maybe we need to test the router too? That is just more complicated and is easy to spot manually I guess afterwards.

azurebrd:
Yeah, I had no idea.  I asked Valerio and he thought that testing through the crud was good.  I don't know how we'd test through the router.

PR #187, opened 2022-07-28
Scrum 1694 2

ianlongden:


azurebrd (approved):
Thanks @ianlongden .  Looks good, I'll merge and try it out.  Good catch removing citation from reference post.

PR #188, opened 2022-07-28
fixed all code to use raw sql and gzip'ed json file before uploading …

sweng66:
…to s3


azurebrd (approved):
That's great, thanks @sweng66 !  Good to see how to generate a zip file in python :)

PR #194, opened 2022-08-02
one off script for deleting duplicate mod_reference_type rows

sweng66:


azurebrd (dismissed):
Thanks @sweng66 I see the SELECT orders by reference_id, and I imagine that it subselects by order it got written in, but it might be better to explicitly subselect by mod_reference_type_id to make sure it finds the earlier ones first and deletes the later ones.  Practically, I don't think it makes a difference, but it might make more sense if we're ever looking at the transactions.  (That a bunch of things that got INSERTed at creation don't have an MRT removed while something from a later update remained).  If you don't think it's worth it, that's fine with me, I realize it's functionally the same thing.

sweng66:
@azurebrd This makes sense!! I fixed it. Thanks.

sweng66:
@azurebrd and will fixed version against prod database when it is ready.

azurebrd (approved):
Awesome, thanks @sweng66 !

PR #195, opened 2022-08-02
adding code to dqm loading program to generate a loading report and send it to slack

sweng66:


azurebrd (dismissed):
Thanks @sweng66 !
I thought the code you're removing in 443-458/479-494 was going to be replaced by the addition of 309-321, but it's commented out.  Is that still in progress, or is that going to happen somewhere else ?  (Totally fine if it's a separate story, I'm not sure what the tickets are)

sweng66:
@azurebrd Good catch! This section is not part of the report (the story is to report any loading errors for PubMed papers) so I commented out this section during testing, I forgot to turn it on before pulling the PR. Will do it now..

Sorry for the confusion!

azurebrd (approved):
Thanks @sweng66 !

PR #196, opened 2022-08-02
XB pmid exclusion listing

pellst:
Ceri requested the PMID exclusion listing for Xenbase


sweng66 (changes requested):
Can you remove the first two lines from the file?  Thanks!

sweng66 (approved):

PR #197, opened 2022-08-03
fixed to avoid adding duplicate mod reference types into database

sweng66:


azurebrd (approved):
It looks like you're always checking whether there's duplicate MRT data for a given reference, each time the DQM load runs.  Probably running it once to clean it, then having a constraint is good enough, but doing it each time just in case seems good.

sweng66:
@azurebrd You are right! Just don't want to introduce any more duplicate rows before having unique constraint in place.  As you said, we  can remove this check at some point.

azurebrd:
Thanks, @sweng66 !

PR #199, opened 2022-08-03
SCRUM-1825

ianlongden:


agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -0,0 +1,168 @@
...
+
+    topic_entity_tag_data = jsonable_encoder(topic_entity_tag)
+
+    if topic_entity_tag_data["reference_id"]:
+        topic_entity_tag_data["reference_curie"] = db.query(ReferenceModel).filter(ReferenceModel.reference_id == topic_entity_tag_data["reference_id"]).first().curie
+        del topic_entity_tag_data["reference_id"]
+
+    props = db.query(TopicEntityTagPropModel).filter(TopicEntityTagPropModel.topic_entity_tag_id == topic_entity_tag_id).all()
+    topic_entity_tag_data["props"] = []
+    prop: TopicEntityTagPropModel
```
azurebrd, line 105: What is this line for ?  I'm not familiar with the syntax, and in the next line prop is each value of props

agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -0,0 +1,168 @@
...
+    db.commit()
+    return {"message": "updated"}
+
+
+def destroy(db: Session, topic_entity_tag_id: int):
+
+    topic_entity_tag = db.query(TopicEntityTagModel).filter(TopicEntityTagModel.topic_entity_tag_id == topic_entity_tag_id).first()
+    if not topic_entity_tag:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND,
+                            detail=f"topic_entityTag with the topic_entity_tag_id {topic_entity_tag_id} is not available")
```
azurebrd, line 164: topic_entityTag : odd casing, not impotant

azurebrd (changes requested):
I think we need a router for the topic_entity_tag_prop, otherwise if topic_entity_tag was created with a qualifier, we won't be able to remove it or update it.  (Edit: we could destroy the whole topic_entity_tag and recreate it with the new qualifiers, but that changes the history.  Maybe it doesn't matter, but it seems more correct to have a full crud on the topic_entity_tag_prop.  We could talk about it)

Edit: There aren't tests for topic_entity_tag_prop or mod_taxon, and that's okay with me.

agr_literature_service/api/schemas/__init__.py
```diff
@@ -81,6 +81,14 @@
     WorkflowTagSchemaRelated,
     WorkflowTagSchemaUpdate)
 
+from agr_literature_service.api.schemas.topic_entity_tag_schemas import (
+    TopicEntityTagPropSchemaCreate,
+    TopicEntityTagPropSchemaShow,
```
azurebrd, line 86: Do we need a TopicEntityTagPropSchemaUpdate ?

agr_literature_service/api/schemas/topic_entity_tag_schemas.py
```diff
@@ -0,0 +1,74 @@
...
+    alliance_entity: Optional[str] = None
+    mod_entity: Optional[str] = None
+    new_entity: Optional[str] = None
+    taxon: int
+    note: Optional[str]
+    date_created: str
+    date_updated: Optional[str]
+    created_by: str
+    updated_by: Optional[str]
+    props: Optional[List[TopicEntityTagPropSchemaCreate]] = None
```
azurebrd, line 74: Should this be 
    props: Optional[List[TopicEntityTagPropSchemaShow]] = None
(Show instead of Create)

agr_literature_service/api/tests/test_018_topic_entity_tag.py
```diff
@@ -0,0 +1,240 @@
...
+db = SessionLocal()
+
+# Add tables/schema if not already there.
+Base.metadata.create_all(engine)
+
+# Exit if this is not a test database, Exit.
+if "literature-test" not in SQLALCHEMY_DATABASE_URL:
+    exit(-1)
+
+fb_mod = None
```
azurebrd, line 36: fb_mod is initialized and updated on mod_create, but doesn't seem part of the test.  This is fine by me.

agr_literature_service/api/crud/topic_entity_tag_crud.py
ianlongden: mypy. It makes mypy happy and editor knows what it is so can check methods etc.

agr_literature_service/api/schemas/__init__.py
ianlongden: I will add a test to see if I can update. Yes, we probably need an update for the props. Also a delete.

agr_literature_service/api/schemas/topic_entity_tag_schemas.py
ianlongden: fixed.

agr_literature_service/api/tests/test_018_topic_entity_tag.py
ianlongden: I may remove this. I just copied and pasted from another initialisation script that needed mods.

agr_literature_service/api/schemas/__init__.py
ianlongden: Nope not really needed. We can update via the TopicEntityTag. I added an example of this in test_018...

azurebrd:
Thanks for explaining and updating things, @ianlongden .

"Yes, we probably need an update for the props. Also a delete."
Ah, okay, I was thinking it'd be better to keep standard with a full CRUD, so we could directly create a topic_entity_tag_prop through its own create endpoint instead of having to patch it through the related topic_entity_tag object it connects to.  I can see that it would work both ways, and it's redundant to add the Create _prop router, but it makes the logic more standard, because when creating other objects that connect to reference, we create them through their create/POST endpoint, instead of thinking about its parent and patching that.

agr_literature_service/api/schemas/topic_entity_tag_schemas.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class TopicEntityTagSchemaCreate(BaseModel):
+    reference_curie: str
+    topic: str
+    entity_type: str
+    alliance_entity: Optional[str] = None
+    mod_entity: Optional[str] = None
+    new_entity: Optional[str] = None
+    taxon: int
```
azurebrd, line 36: I think taxon should be a string, because we'd be storing the NCBI taxon id, which is a curie.  The only value we'll be using for the current ticket is NCBITaxon:559292

ianlongden: AHH! I was expecting just the number bit i.e. 559292. That is what is normal for taxon ids. But I can change it, as it looks like you want to store the curie. Other option is to strip off the 'NCBITaxon:' but I guess if the A-team is returning the string version, then we should store as is. I presume this is due to them storing this in a xref type table.

azurebrd: I think that because it's an ontology term it's keying off the full curie.  You're right though, at Caltech we just store the numeric part of the taxon curie in the taxon table.  I'll ask them to make sure.  Also, editing the comment to mention we need mod_taxon routers, and I would prefer a crud so we don't have to write sqlalchemy to add/delete/patch, but also need a get endpoint that takes a mod to return a list of species, and I'm not sure what that should be called.

agr_literature_service/api/models/mod_taxon_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+    mod_id = Column(
+        Integer,
+        ForeignKey("mod.mod_id"),
+        unique=False,
+        nullable=False
+    )
+
+    # taxon_id
+    taxon = Column(
```
azurebrd, line 33: From the conversation with Sierra and Olin
Sierra : `so your column name could be "taxon_id" (that could also be the slot name for the class).
in our alliance model, we have the slot "taxon" which fulfills this purpose already`
and
Olin : `yes we store the full curie in the curie field and when linking typically it would be taxon_id which would point to the curie field`
I think I was confused by her mentioning the slot "taxon" but the column name being "taxon_id".  I guess they're saying the column should be taxon_id.  Up to you whether we should rename the column, so far we're using _id to mean an internal database ID, right ?
They both seem to agree that it should store the curie String() though

ianlongden: It is now taxon. We can change it later if it becomes important.

agr_literature_service/api/schemas/topic_entity_tag_schemas.py
ianlongden: Lets add mod_taxon router stuff as a separate story later. Presently this is not needed and feels a bit like a waste of time for something so small at the moment. I will let the PO set the priority etc...

agr_literature_service/api/models/mod_taxon_model.py
azurebrd: Sounds good

agr_literature_service/api/schemas/topic_entity_tag_schemas.py
azurebrd: Sounds good

azurebrd:
topic still doesn't show on reference GET https://dev4003-literature-rest.alliancegenome.org/reference/AGR:AGR-Reference-0000000001  Also doesn't show in swagger schema for POST / PATCH

topic_entity_tag POST works with and without props, but patch to add qualifiers doesn't with 500 error (works to patch without props)
{
  "note": "a note",
  "props": [ { "qualifier": "ATP:qualifier1" } ]
}

/topic_entity_tag_prop/{topic_entity_tag_prop_id}  PATCH  schema example has
{
  "qualifier": "string",
  "topic_entity_tag_prop_id": 0
}
with topic_entity_tag_prop_id instead of topic_entity_tag_id
Trying to patch it with topic_entity_tag_prop_id 1 in the endpoint and passing
{
  "qualifier": "patch prop through PATCH",
  "topic_entity_tag_id": 2
}
gives a 202 success, but does not update the value

agr_literature_service/api/schemas/reference_schemas.py
```diff
@@ -137,6 +138,7 @@ class ReferenceSchemaShow(BaseModelShow):
     open_access: Optional[bool] = None
     citation: Optional[str] = None
     workflow_tags: Optional[List[WorkflowTagSchemaShow]] = None
+    topic_entity_tag: Optional[List[TopicEntityTagSchemaShow]] = None
```
azurebrd, line 141: topic_entity_tags instead of  topic_entity_tag.  But this also returns a '' instead of a [] so something else is wrong

azurebrd (approved):
Thanks @ianlongden !

PR #200, opened 2022-08-03
fixed the 'mod_reference_types' to 'mod_reference_type'

sweng66:


azurebrd (approved):
Thanks for tracking this down, @sweng66 !

PR #206, opened 2022-08-09
Tidy test

ianlongden:
Tidying scripts to just use one initialisation script in a new files utils.


azurebrd (approved):
Great, thanks @ianlongden !

PR #209, opened 2022-08-12
script for adding missing errata/correction papers into mod_corpus_as…

sweng66:
…sociation


azurebrd (commented):
Hi @sweng66 .  Wow this correction stuff is complicated.  (I'm not sure I follow everything)

1- I think they only want to transfer values that are “in Corpus”, which I think is represented in the database by the corpus column being true (false meaning outside corpus, and null meaning needs review).  So the query on 168 should probably restrict to WHERE corpus = True (or whatever the syntax is for booleans).

2- The comment and correction population from pubmed was recursive, so for every paper we downloaded all the corrections, and for each of those, we downloaded their corrections.  At some point something was 13 levels deep, and Ceri thought we were getting too much junk, so we removed one of the categories, and I think we ended up with something 9 levels deep.  I think this script runs once, and in line 64, when both from + to are not in mca, they're skipped.  But in subsequent runs, they might have been associated to their own correction, so it should probably be run recursively, or at least 9 times.  Or maybe just once and see afterward how many references still don't have a connection.  Maybe it'd be good to separately track the references that don't have an MCA connection at all, those being the ones that need a connection, and track the ones that have a corpus == null, those being ones that need review that came from a mod search.   I can see there's different ways to go about it.  I'd have probably written it to start from the references that don't have an mca value at all, and then looped through their corrections to see if any corrections had an MCA == True.  I'm sure what you're doing also works though.

3- In lines 79-92, I think you're going through existing corrections to get all possible mca from and to, and then if a main paper has an MCA value the associated one doesn't, you add the connection.  That makes sense, but I'm not sure it's what Ceri wants, so I'd check in with her.  I would think that if the associated paper is already associated with some mod as corpus == True, that'd be good enough, and other mods don't need to be associated even if their parents are.  But here I really don't know, it makes sense that they'd want to have papers associated.  But maybe some mods don't want to have a bunch of corrections associated with them ?  And they are only okay with it happening so we don't have any papers floating unassociated with any mod ?  I really don't know.  It makes sense that this could also be a very good clean up step that I wouldn't have done by starting with unassociated papers.

4- Kind of same thing with 95-108.  I can see that being good cleanup, or too much info, and would check with Ceri.  Line 100 says that the update is happening if the main paper's corpus is True, which sounds good, but I couldn't see how it was filtered to just True values.  If that's already happening, then what I said in point 1 might not be right.

5- I'm not sure which is the main paper (from / to), but you seem to know, so that's good.  I'll assume 110-117 makes sense, unless it's like points 3 and 4.

For background, when we first populated the DQM data, they wanted us to loop recursively through the corrections, so we did.  But they wanted that just to get legacy papers that might have been missed.  Now for ongoing pubmed searches and updates, they don't care about corrections, because the usual pubmed search is supposed to be good enough.  It might be because of that that I'm not sure if Ceri would want those connections that might be happening in points 3-5.  Sorry that this is more complicated than starting fresh, if we only had pubmed papers that came from DQMs and were creating their comments/corrections, it would have probably been easier to know what needed to get associated.

Thanks @sweng66 !

azurebrd (dismissed):
I don't know what the source is, so I can't tell what it's really doing, but the sqlalchemy makes sense.  If it's good for curators, sounds good to me, @sweng66 :)

sweng66:
@azurebrd Thanks for checking the code. I got a list of reference_id and mod_id pair from rdsprod loading log file (exactly same as rdsdev loading file) so we know this is the list of mod association rows just added for comments_and_corrections yesterday. And set the mod_corpus_sort_source to 'automated_alliance' for these newly added rows.

azurebrd (approved):
Great, thanks @sweng66 !

PR #212, opened 2022-08-17
feat(dev_app): added vim and emacs

valearna:


sweng66 (approved):
This is great! Thanks!

PR #216, opened 2022-08-23
adding code to download the xml files again for missing pmids

sweng66:


valearna (approved):
Looks good, thanks @sweng66! Maybe adding a retry with exponential timeout to the download function would have been easier to manage, but your approach is totally fine

PR #217, opened 2022-08-23
Scrum 1887 - Fixed automated fields for AuditedObject

valearna:
- Fixed default values for date_updated and date_created by using function definitions as default values instead of function calls (which were cached when the python module was imported)
- Automatically set created_by and updated_by to currently logged-in okta user, and use default_user when using sqlalchemy directly (for the following tables: author, editor, mod, mod_corpus_association, reference, resource, topic_entity_tag, topic_entity_tag_prop and workflow_tag, mod_taxon)
- Using onupdate to automatically set updated_by without the need to call utils functions inside the cruds
- Simplified schemas by using inheritance from AuditedObjectModelSchema for the 4 automated fields (date_updated, date_created, updated_by, created_by) - all fields are optional in the schema but we should discuss if we need them to be required after we populate the fields for older entries in the db
- Generated alembic version file for db changes
- Current topic_entity_tag allows overwriting the 4 audited fields manually and we fixed bugs that were preventing this in the crud but we should consider whether we want to remove this feature in the future
- Added a function to create the default_user when the db is initialized (if not already present)


valearna:
@sweng66 is currently working to remove code that manually sets date_updated for references in the pubmed update program as this field is now set automatically by using the sqlalchemy onupdate option

azurebrd (approved):
Thanks @valearna and @sweng66

PR #218, opened 2022-08-24
adding reports for the papers in ABC, but in MOD dqm submission

sweng66:


azurebrd (dismissed):
Cool how the width formatting works, @sweng66 .

In line 700, why does that only happen when i < 6 ?

In line 714 isn't log_url always something because line 711 appends log_file to it ?  Maybe move line 711 inside 714 so 714 only refers to the ENV variable ?

sweng66:
Thanks for looking over the code @azurebrd !

1. "In line 700, why does that only happen when i < 6 ?"

We only want to show the first five rows in the report, but curators can go to the log file to see the full list

2. "In line 714 isn't log_url always something because line 711 appends log_file to it ? Maybe move line 711 inside 714 so 714 only refers to the ENV variable ?"

Good catch!!  I fixed that. Thanks

azurebrd:
Ah, cool, thanks @sweng66 !  
Yikes, I've no idea why all the tests are failing.  I don't see anything commits in main that didn't pass tests.  Sometimes things go wrong on github, and restarting the tests works.  I've only done that by making a meaningless commit/push, but there might be a way to restart the tests directly.

azurebrd (approved):
Thanks @sweng66 !

PR #219, opened 2022-08-25
fix(fastapi_okta): using new agr_fastapi_okta repository

valearna:
the old fastapi_okta fork was deleted and we lost git history so we decided to create a new repository with Aoji's files


azurebrd (approved):
Thanks @valearna @sweng66 and @FB-ping

PR #220, opened 2022-08-25
fix(db_models): added ondelete cascade tags

valearna:


azurebrd (approved):
Thanks @valearna , @sweng66 , @ianlongden , @FB-ping

PR #223, opened 2022-08-30
Fixed to use newly updated fastapi_okta module

sweng66:


sweng66:
@azurebrd oneoff script looks good! Thanks.

PR #225, opened 2022-09-01
Scrum 1895 - New unit tests using fastapi testing framework

valearna:
- created new unit tests based on fastapi testing framework, which lets you use a test api object to make calls to the endpoints and thus test the routers and not only the crud functions
- removed previous unit test files from api code
- created new tests dir structure in the root dir
- added pytest fixtures to wipe the db before and after each test (can be controlled through a new env var TEST_DATABASE_DELETE)
- added fixtures to create test objects in the database
- each test function is now independent and works on its own test objects


sweng66 (approved):
@valearna Looks great! Thanks.

tests/api/test_mod_corpus_association.py
```diff
@@ -0,0 +1,84 @@
...
+TestMCAData = namedtuple('TestMCAData', ['response', 'new_mca_id', 'related_ref_curie'])
+
+
+@pytest.fixture
+def create_test_mca(db, auth_headers, create_test_reference, create_test_mod): # noqa
+    print("***** Adding a test mod-corpus association *****")
+    with TestClient(app) as client:
+        mod_response = client.get(url=f"/mod/{create_test_mod.new_mod_abbreviation}")
+        mod_abbreviation = mod_response.json()["abbreviation"]
+        new_mcc = {
```
azurebrd, line 22: new_mca would probably be clearer, but it makes sense

tests/api/test_mod_corpus_association.py
```diff
@@ -0,0 +1,84 @@
...
+            assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    def test_create_mca(self, create_test_mca):
+        assert create_test_mca.response.status_code == status.HTTP_201_CREATED
+
+    def test_show_by_reference_mod_abbreviation(self, create_test_mca):
+        with TestClient(app) as client:
+            test_mca_response = client.get(url=f"/reference/mod_corpus_association/{create_test_mca.new_mca_id}")
+            test_mca_abbreviation = test_mca_response.json()["mod_abbreviation"]
+            response = client.get(url=f"/reference/mod_corpus_association/"
```
azurebrd, line 45: Is this a shortcut for doing 3 gets in a row, then the assert on line 48 requires all 3 to have the same status_code ?
The f-string on 46 and 47 don't start with / like the other client.get(url=f"  and I'm not sure if it means something else or it's the syntax for multi-request client.get

tests/api/test_workflow_tag.py
```diff
@@ -0,0 +1,127 @@
...
+                  "workflow_tag_id": "ont1",
+                  }
+        response = client.post(url="/workflow_tag/", json=new_wt, headers=auth_headers)
+        yield TestWTData(response, response.json(), test_reference.json(), test_mod.new_mod_id,
+                         test_mod.new_mod_abbreviation)
+
+
+class TestWorkflowTag:
+
+    def test_get_bad_ref_ont(self):
```
azurebrd, line 32: ref_ont is kind of confusing, why not workflow_tag ?  It works though, so it's fine.  Same at 66-70 and the new method definitions.

tests/api/test_ref_cmmnt_and_crrctn.py
```diff
@@ -0,0 +1,115 @@
...
+        assert rcc_obj.reference_comment_and_correction_type == "CommentOn"
+
+    def test_patch_rcc(self, db, test_ref_cc, auth_headers): # noqa
+        with TestClient(app) as client:
+            # swap to and from and change correction type
+            xml = {'reference_curie_from': test_ref_cc.ref_curie_to,
+                   'reference_curie_to': test_ref_cc.ref_curie_from,
+                   'reference_comment_and_correction_type': "ReprintOf"
+                   }
+            response = client.patch(url=f"reference_comment_and_correction/{test_ref_cc.new_rcc_id}",
```
azurebrd, line 72: should this have a / between url=f" and reference_comment_and_correction ?  same with lines 83 85 86

tests/api/test_topic_entity_tag.py
```diff
@@ -0,0 +1,218 @@
...
+                "topic": "Topic1",
+                "entity_type": "Gene",
+                "taxon": "NCBITaxon:1234"
+            }
+            # No Entities
+            response = client.post(url="/topic_entity_tag/", json=xml, headers=auth_headers)
+            assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
+
+            # More than one Entity
+            xml2 = copy.deepcopy(xml)
```
azurebrd, line 80: xml is already 422.  Might it be better to start with a working xml, test that it works, then deepcopy that working xml and modify it to break in different ways to test that it's breaking because of the changes, instead of the first one broke, so all copies of it break regardless of different reasons for breaking ?

tests/api/test_topic_entity_tag.py
```diff
@@ -0,0 +1,218 @@
...
+            # No species
+            xml5 = copy.deepcopy(xml4)
+            del xml5["taxon"]
+            xml5["reference_curie"] = test_reference.json()
+            response = client.post(url="/topic_entity_tag/", json=xml4, headers=auth_headers)
+            assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
+
+    def test_patch_with_props(self, test_topic_entity_tag, test_reference, auth_headers): # noqa
+        with TestClient(app) as client:
+            # change the reference
```
azurebrd, line 108: I'm not sure this is testing changing the reference, is the reference_curie being patched the same as the original and then testing that the response is still the original/patch ?

azurebrd (dismissed):
The namedtuple is so much more readable.
I had some comments, but I think everything's really fine.  Some of the topic_entity_tag and workflow_tag tests we probably won't want once Pinglei finishes what he's doing.  Hopefully we'll merge this in, and he'll remove unwanted tests as part of the work he's doing.  Possibly some workflow_tag tests.  
Later we should add more workflow and tet tests, but with the rework of tet coming, maybe it's not worth doing it for tet yet.

tests/api/test_mod_corpus_association.py
valearna: I'll change it to mca, probably a typo

tests/api/test_workflow_tag.py
valearna: the old unit tests used _ont but I agree that it is confusing. I'll change them to _wt

tests/api/test_ref_cmmnt_and_crrctn.py
valearna: I will add the missing "/". I think that it should be there, at least to be consistent with the other calls

tests/api/test_topic_entity_tag.py
valearna: I copied this from the old unit tests, but what you are saying makes sense. I'll rework this test following your suggestion

tests/api/test_topic_entity_tag.py
valearna: I'm not sure about this. I'll take a look and modify the test if needed. I tried to keep the same logic as the original unit test file but maybe something got lost here

tests/api/test_topic_entity_tag.py
valearna: We looked at this together and realized that the tests are ok, even though they are a bit confusing. We decided to keep the code as is

tests/api/test_topic_entity_tag.py
valearna: we fixed this issue together by using test_reference2

azurebrd (approved):
Great, thanks @valearna

PR #227, opened 2022-09-02
adding code to set global user

sweng66:


azurebrd (approved):
Great, thanks @sweng66 !

PR #228, opened 2022-09-02
fix(continuum): tracking inserts with operation type 0 instead of 1

valearna:
multiple calls to enable_versioning() now result in a single call to make_versioned

- avoids tracking changes multiple times. Calling make_versioned more than once was causing continuum operation types to be always 1 instead of 0 because continuum thought that if the same object is inserted twice in the same transaction there's probably a delete before the last insert and it was tracking the resulting operation as an update
- added call to enable_versioning to all models

### DIFF
--- a/agr_literature_service/api/database/versioning.py
+++ b/agr_literature_service/api/database/versioning.py
@@ -2,7 +2,12 @@ from sqlalchemy_continuum import make_versioned
 from sqlalchemy_continuum.plugins import PropertyModTrackerPlugin
 from agr_literature_service.api.continuum_plugins import UserPlugin
 
+already_called = False
+
 
 def enable_versioning():
-    user_plugin = UserPlugin()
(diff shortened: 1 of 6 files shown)


sweng66 (commented):
@valearna What does the change mean in versioning.py file? everything else looks good.

valearna:
@sweng66 the new code in versioning.py makes sure that make_versioned is called only once per session. All models are calling enable_versioning() but only the first call now results in make_versioned being called, and so there's now only one tracker for sqlalchemy_continuum

sweng66 (dismissed):
@valearna Cool! Thanks.

valearna:
added a decorator to run functions only once instead of using global vars. This should make the code easier to understand

sweng66 (approved):
Cool! Thanks @valearna

valearna:
thanks @sweng66!

azurebrd:
That's great, thanks @valearna and @sweng66 !

PR #229, opened 2022-09-03
Scrum 1853 workflow tag

FB-ping:
modify topic_entity_tag and workflow_tag


azurebrd (changes requested):
Hi @FB-ping 
We definitely don't want the .env.kenichi nor the changes to the alembic.ini  If possible when doing the github add, just add the files that you want instead of * 

Do you know if the change to the workflow patch would allow sending in an empty string for the mod ?  Previously that was broken.  One could create it with blank (correct), but could not patch it to blank.

I'm pretty sure we want the schemas to inherit from AuditedObjectModelSchema instead of BaseModel on line 7 and 33/42 and 57/70 of topic_entity_tag_schemas.  I'm not sure about line 13 for the props.  Same for workflow base/audited line 30/38.  Did you keep what happened from merging main into your branch ?  Valerio's Audited stuff takes care of updating created_by updated_by date_updated date_created, so we should be using that for these 3 tables, instead of BaseModel.

Why is it now a TopicEntityTagSchemaShowRelated instead of TopicEntityTagSchemaRelated ?  

Thanks !

valearna:
@azurebrd @FB-ping the new schemas look good to me. All updates and posts should not inherit from AuditedObject since we don't want to set those fields from the API but they are managed only by SQLAlchemy. It would be useful to have a live demo of these changes and review all endpoints together

sweng66:
@valearna Sounds good to have a live code review session to go over everything together

FB-ping:
@azurebrd  1. sorry, I uploaded the .env.kenich by mistake, removed now. 2. I fixed the workflow_tag so it can have "" value for mod_abbreviation when use patch  3. as @valearna for create/patch, we should inherit from BaseModel instead of AuditedObject since we don't want to set those fields from the API but they are managed only by SQLAlchemy  4. rename TopicEntityTagSchemaRelated to TopicEntityTagSchemaShowRelated so we know that we will use this to show data by parents(eg. via reference to show topic_entity_tag data), and use TopicEntityTagSchemaShow when use to show data for itself.    not sure what's name convention, I can change back to TopicEntityTagSchemaRelated if we have name convention.

sweng66:
@azurebrd I also want to know the naming convention. agree! @valearna  a demo of the models and tests would be great

valearna:
I would prefer to keep the same naming convention we have in all other schemas, so "Related" instead of "ShowRelated"

azurebrd:
Thanks @FB-ping glad to see that test passes :)  I'll defer to @valearna about the other changes, but I imagine it's all good

valearna (changes requested):
remove FB credentials from alembic.ini

azurebrd:
Thanks @FB-ping :)

valearna (approved):

azurebrd (approved):

PR #230, opened 2022-09-08
fix model of topic_entity_tag and workflow_tag

FB-ping:
fix reference delete function with topic_entity_tag and workflow_tag data


azurebrd (approved):
Thanks @FB-ping !

PR #231, opened 2022-09-09
Scrum 1957 - Refactoring lit_processing and functional tests

valearna:
- restructured lit_processing module - created new code structure, separated code from tmp files using by pipeline, deleted unused files and moved oneoff scripts to separate folder, renamed files to make them easier to identify
- updated functional tests to work directly from python script and not as a collection of shell scripts
- removed dependency on api from lit_processing and used sqlalchemy directly and set global user_id for scripts (Shuai's work)
- new env var for FMS_RELEASE version
- new function to initialize db (create tables if not exist)
- added base_dir variable to functions that can work from dir different than XML_PATH
- refactored main functions in scripts to call it both from command line and import it from other modules
- added function to create tmp dir if not exists
- added cleanup script to remove tmp files
- removed dependency on api in test_runner


azurebrd (approved):
Thanks @valearna and @sweng66 !

PR #232, opened 2022-09-12
added a check for json file to make sure it exists before processing + also added a missing line to add mca row into database

sweng66:


azurebrd (approved):
Great, thanks @sweng66 !

PR #236, opened 2022-09-22
SCRUM-2046 python test of alpha-mati access.  Use in SCRUM-1976 and t…

azurebrd:
…hen could be removed from repo


sweng66 (approved):
Looks good! Thanks @azurebrd

PR #238, opened 2022-09-28
create two new utils modules for db-retrieval/update functions and for slack reporting functions

sweng66:


azurebrd (approved):
Thanks @sweng66 !

PR #240, opened 2022-10-03
Scrum 2097 - dqm ingest refactoring

valearna:
- split up aggregate_dqm_reference from parse_dqm_json_reference.py into subfunctions and refactored code
- created ReportWriter class to take care of report/logs
- moved constants ouside main function
- created memoized class in utils and moved execute_once into shared utils between api and lit_processing
- fixed bugs with dqm_ingest code and updated test data accordingly


azurebrd (approved):
Thanks @valearna

PR #242, opened 2022-10-03
Moving db_reading functions from data exporting program to db_read_utils..

sweng66:
and moving db_updating functions from PubMed update program to db_write_utils.
also making 'update_authors' function work for both DQM loading and PubMed update.


sweng66:
@valearna Thanks for checking the code out!

azurebrd (approved):
Great, thanks @sweng66 !

PR #243, opened 2022-10-04
Scrum 2097 (continued) - Created "Reference" class 

valearna:
- Moved all functions that modify references to the new Reference class
- A ref object can be used as a dictionary (as before) but it now contains class functions to modify its internal state


sweng66 (approved):
Looks great! @valearna

PR #250, opened 2022-10-11
New Reference/Resource curie ID minting related code update

sweng66:
1. added one-off scripts for updating Reference / Resource curies for existing References and Resources in the database
2. fixed the new reference/resource curie generating functions to get ID one by one from ID minting Web Service and moved the functions from api/crud/reference_crud.py and resource_crud.py to global_utils.py
3. fixed  api/lit_processing code to call the functions from global_utils.py
4. fixed the unit tests for api/search
5. fixed the okta_utils.py to check the okta token file to make sure it is less than one day old if the file exists. If the file is older than a day, generate a new okta token and store it in the okta file.


agr_literature_service/global_utils.py
```diff
@@ -40,3 +48,56 @@ def wrapper(*args, **kwargs):
...
+
+    ### it is only for testing purpose
+    # db_session = db
+    # if db is None:
+    #    db_session = create_postgres_session(False)
+    curie_start = "AGRKB:102"
+    rs = None
+    if subdomain == 'reference':
+        curie_start = "AGRKB:101"
+        rs = db.execute("SELECT curie FROM reference order by reference_id desc")
```
azurebrd, line 78: @sweng66 is it any more efficient to add a LIMIT 1 to only get one ?

agr_literature_service/lit_processing/oneoff_scripts/update_reference_curie.py
```diff
@@ -0,0 +1,81 @@
...
+        db_session.commit()
+
+    db_session.close()
+
+
+def get_next_curie(curie, counter):
+
+    curie_len = len(curie)
+    counter_len = len(str(counter))
+    next_curie = curie[0:curie_len - counter_len] + str(counter + 1)
```
azurebrd, line 56: @sweng66 I just want to check I understand this.  get_first_curie returns res_json['first']['curie'] which is a curie string, and that gets passed here as `curie`.  so curie_len is the amount of characters in the string, and counter_len is the amount of characters in the counter number.  Then `curie[0:curie_len - counter_len]` splices the curie string to have the same beginning, and then appends counter+1, which could in theory be one digit, and then 10 curies later it would be two digits, and it works out because the counter variable is getting updated in the update_curies for loop ?

agr_literature_service/global_utils.py
sweng66: Good idea! Initially I thought this is just for unit tests so it only has a few references in the test db anyway, but I think it is a good practice to do so! Thanks for the suggestion! I will fix it @azurebrd

agr_literature_service/lit_processing/utils/okta_utils.py
```diff
@@ -60,9 +62,14 @@ def get_authentication_token():  # pragma: no cover
     okta_file = base_path + 'okta_token'
     token = ''
     if path.isfile(okta_file):
-        with open(okta_file, 'r') as okta_fh:
-            token = okta_fh.read().replace("\n", "")
-            okta_fh.close
+        one_day_ago = datetime.now() - relativedelta(days=1)
```
azurebrd, line 65: 1 day is probably fine.  If okta is really strict about 1 day, we might have issues if there's some lag somewhere, but hopefully it'll be fine, and we don't currently have any processes that take 24+ hours

docker-compose.yaml
```diff
@@ -86,6 +86,7 @@ services:
       OKTA_CLIENT_ID: "${OKTA_CLIENT_ID}"
       OKTA_CLIENT_SECRET: "${OKTA_CLIENT_SECRET}"
       ENV_STATE: "${ENV_STATE}"
+      ID_MATI_URL: "${ID_MATI_URL}"
```
azurebrd, line 89: @sweng66 I don't know which services need this value, but I'll defer to you and @valearna

azurebrd (approved):
That's great, thanks @sweng66 !  Some comments in there, but I think it's good

agr_literature_service/lit_processing/utils/okta_utils.py
sweng66: Yes, you are right! hopefully everything will work out fine. If not, we can fix again. =)

agr_literature_service/lit_processing/oneoff_scripts/update_reference_curie.py
sweng66: yes, correct!

docker-compose.yaml
sweng66: API and crontab all need ID_MATI_URL for getting the new curie.

agr_literature_service/lit_processing/oneoff_scripts/update_reference_curie.py
azurebrd: Thanks @sweng66 !

azurebrd:
Thanks for the responses, @sweng66 !

azurebrd (dismissed):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

valearna (approved):

PR #253, opened 2022-10-13
fix(mod_ref_type): insert ref_mod_reftype entry only if mod_reftype entry exists in the db

valearna:
…lready exists


azurebrd (approved):
Thanks @valearna !

sweng66 (approved):
No more mod_referencetype_id = None in the reference_mod_referencetype table.
Thanks for the quick fix @valearna

sweng66:
It also adds new SGD referencetype into referencetype/mod_referencetype/reference_mod_referencetype tables so it is good! @valearna  Thanks!

PR #254, opened 2022-10-13
update lit_processing programs to populate/update date_published_start/end columns

sweng66:
1. fixed xml_to_json.py to add date_published_start/end fields
2. fixed dqm loading, PubMed search, PubMed update to populate / update these two new columns and to report any un-parsable date_published (sending reports to slack)
3. added a check to make sure returned curie(s) is not None before extending the new_curie_list in post_reference_to_db.py
4. added code section to extract values for these two new columns if date_published has a value but date_published_start/end don't (mainly for unit tests)
5. added the duplicate DOI report back


azurebrd (dismissed):
That's great @sweng66 !  Everything that it does makes sense.  I don't have a whole picture of what it should all be doing or what needs to be done, but I defer to you on that, so, seems good !

sweng66:
Thanks! @azurebrd

azurebrd (approved):
haha, missed that typo too.  the variable name made sense :P

sweng66:
👍 @azurebrd

valearna (approved):

PR #255, opened 2022-10-13
Refactor idl

ianlongden:
Round 1 complete. Still need to look at the processing of the initial dqm file and probably a little more tiding up but it does what is expected now. Will become easier with the constraints to come but for now this will do.


agr_literature_service/lit_processing/data_ingest/post_resource_to_db.py
```diff
@@ -49,9 +53,174 @@
...
+    Parse the json file and load the data into the db after some remapping of keys etc.
+    i.e.
+    base_input_dir = "/user/bob/")
+    json_path = "sanitized_resources/"
+    post_resources(db, json_path, 'ZFIN', base_input_dir)
+
+    reads the file
+    /user/bob/sanitized_resource_json_updates/RESOURCE_ZFIN.json"
+
+    If base_input_dir not defined environment value of XML_BASE used.
```
azurebrd, line 222: I think it's XML_PATH instead of XML_BASE, might be confusing to someone reading this later

azurebrd (commented):
I'm assuming the logic hasn't change, like in update_resource being moved to its own function instead of the loop.  I'm not checking carefully that it doesn't have added or removed logic that makes sense.  (same with process_editors, process_cross_references, remap_keys_get_new_entry, process_resource_entry)  From the tests, I'm sure some logic has changed, but I'm not sure what to focus on, please let me know if there's particular functions that should get more attention.

This is basically an 'Approve' of the PR, but I think @valearna and @sweng66 should look this over before merging.  Might be good to go over it together.

agr_literature_service/lit_processing/data_ingest/post_resource_to_db.py
ianlongden: Sorted thanks.

ianlongden:
Sure, I think looking at the tests should be a better judge of what is good/bad with this many changes it is REALLY hard to see what has changed.

sweng66:
Looks good to me! Thanks @ianlongden 

xref_ref: Dict = {}
xref_ref_valid = {}

Was wondering about these hashes.. What is the difference between these two? Are they not the same thing?

azurebrd:
@sweng66 
xref_ref: Dict = {} has all xref to ref mappings, even obsolete ones, to prevent creating the same xref curie, or something.
xref_ref_valid = {} has only the valid xref to ref mappings.  I forget why sometimes you want to map things to valid xrefs only, but I think there were cases where you needed that.

sweng66:
@azurebrd I understand this part, but thanks for explaining. I was actually wondering about syntax..

a = {}
a: Dict={}

azurebrd:
Oh, sorry, @sweng66 .  I have no idea.  mypy sometimes complains about typing and other times it doesn't.  Maybe that was it.  I'll defer to @ianlongden and @valearna

sweng66:
Totally ignore it. Not a big deal. Thanks @azurebrd

azurebrd (commented):
@ianlongden What is this script for ? [agr_literature_service/lit_processing/oneoff_scripts/load_dqm_resource.py](https://github.com/alliance-genome/agr_literature_service/pull/255/files#diff-c83bf0e5e26dea2a10ce2d0394056c58ca1180e4a394dc1d61b2657c9ed3f36a) 

I've lost track of what's part of pipelines.  I think oneoff/ is for things that run once, but the original loading of dqm resources already happened, so only sort_dqm_resource_updates.py would be run again in the future.   I guess this is a replacement for parse_dqm_json_resource.py ?

ianlongden:
Very strange...
FAILED tests/lit_processing/data_ingest/utils/test_db_write_utils.py::TestDbReadUtils::test_dqm_update_functions
This is for references though which I did not change?

ianlongden:
Hmm, no changes made,  I reran the tests and all is good? Something is not quite right here. Lets see if it happens again.

azurebrd:
> Hmm, no changes made, I reran the tests and all is good? Something is not quite right here. Lets see if it happens again.

@ianlongden Valerio found that some tests were inserting then reading things, but not querying them in order, so some things would randomly be written in a different order than in the source, so a test would fail.  He fixed the one he found, but we assume other tests were written the other way, and we'd have to look at all the tests, or see if they randomly fail.

PR #256, opened 2022-10-13
Scrum 2156 - new script to populate new mod ref type table from old table

valearna:
- will need another PR to delete the old table with alembic


azurebrd (approved):
Thanks @valearna !  Looks good.  I think you said the alembic constraints would get generated during the removal of the old model during cleanup.  So a heads up to people that might do alembic work that it will probably try to add these constraints.

sweng66 (approved):
Looks good! @valearna

sweng66 (approved):

PR #259, opened 2022-10-15
Mod ref type patch hotfix

valearna:
- returning error codes from mod ref type patch api when passing invalid label and/or mod
- now allowing to patch to existing mod_referencetype entries only


sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):

PR #262, opened 2022-10-21
Scrum 2108 - pubmed dates as date objects instead of strings in elasticsearch

valearna:
- changed mappings to parse pubmed dates (date_arrived_in_pubmed and date_last_modified_in_pubmed) as es dates and not as strings
- automated debezium to simplify restart


sweng66 (approved):
Looks good! although I didn't understand fully, but got the point. Thanks @valearna

azurebrd (commented):
That's great @valearna !  (Good to see how to do those new things for deployment and initialization and remapping values and handling null values)

PR #264, opened 2022-10-25
fixes to add unique constraints to cross_reference table + set curie_prefix as a non-nullable columns

sweng66:
and reset primary key for cross_reference_version table 'cross_reference_id'.


azurebrd (approved):
Thanks @sweng66 and @valearna !

PR #270, opened 2022-10-28
fix(needs_review): cross_reference_schema has changed, search_crud ne…

azurebrd:
…eds additional values.


sweng66 (dismissed):
THe change makes sense !

valearna (approved):

PR #273, opened 2022-10-30
Fixed a bug in XREF curie for Resource loading

sweng66:


azurebrd (approved):
That's great, thanks @sweng66 .  I'm not merging this so @ianlongden can look it over first.

ianlongden:
AH! I thought the prefix and identifier were supposed to be split now, not just adding a prefix aswell. Duplication of data but if this is what was wanted then this is all good.

valearna (approved):

azurebrd:
@ianlongden Yeah, it's still a curie field, so we have the whole thing.  I guess if we converted it to an identifier field we'd have the non-prefix part of the curie, and then we could have some other field that could be the aggregation ?  I don't know if it's possible to have a fake column that means return this + that column.  I don't know how much we're still trying to be really tight with the LinkML, so maybe we could move away from having a curie in the future.

PR #275, opened 2022-11-01
Scrum 2199 - delete old mod reference type table

valearna:
deleted models and created alembic revision file


sweng66 (dismissed):
Looks good! Thanks @valearna

valearna:
thanks @sweng66, I had to remove the oneoff script to populate the new tables from the old one since it was importing the old model. Can you re-review the PR?

azurebrd (approved):
Thanks @valearna !

PR #278, opened 2022-11-03
referencefile pdf_type and referencefile_mod mod_id should be nullabl…

azurebrd:
…e.  changed model and made new alembic version file


valearna (approved):

sweng66 (approved):
Looks good! Thanks @azurebrd

PR #279, opened 2022-11-05
Scrum 2260 - new API for referencefile and referencefile_mod association

valearna:
This PR adds API endpoints with full CRUD operations for referencefile and referencefile_mod tables. In addition, reference show now returns the list of referencefiles of each reference and referencefile show returns the list of referencefile_mods. To make it easier to retrieve the list of referencefile objects of each reference, the PR adds a back_populate relationship (many-to-one) for referencefile into the ReferenceModel


sweng66 (dismissed):
Looks great! Thanks @valearna 

There is one place In crud/referencefile_crud.py:

def destroy(db: Session, md5sum_or_referencefile_id: str):
    referencefile = read_referencefile_db_obj_from_md5sum_or_id(db, md5sum_or_referencefile_id)
    db.delete(referencefile)
    db.commit()

Do we need to add a check?  for example,

if not referencefile:
        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND,
                   detail=f"referencefile with referencefile_id or md5sum {md5sum_or_referencefile_id} not found")

 db.delete(referencefile)

Same for destroy function in crud/referencefile_mod_crud.py.

sweng66:
Also In crud/referencefile_crud.py:

line39-40:

referencefile_dict["reference_curie"] = db.query(ReferenceModel.curie).filter(
        ReferenceModel.reference_id == referencefile_dict["reference_id"]).one()[0]

Will this complain if it returns None or it won't return None?

azurebrd:
@sweng66 

> def destroy(db: Session, md5sum_or_referencefile_id: str): referencefile = read_referencefile_db_obj_from_md5sum_or_id(db, md5sum_or_referencefile_id) db.delete(referencefile) db.commit()
> 
> Do we need to add a check? for example,
> 
> if not referencefile: raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"referencefile with referencefile_id or md5sum {md5sum_or_referencefile_id} not found")
> 
> db.delete(referencefile)

I think because `read_referencefile_db_obj_from_md5sum_or_id(db, md5sum_or_referencefile_id)` happens first, in that function the `if not referencefile: raise` happens, and I think Valerio said that it would basically exit out because of the exception, so the rest of the destroy method doesn't get executed

azurebrd:
@sweng66 

> referencefile_dict["reference_curie"] = db.query(ReferenceModel.curie).filter( ReferenceModel.reference_id == referencefile_dict["reference_id"]).one()[0]
> 
> Will this complain if it returns None or it won't return None?

I think because reference_id is required in the table, and it's a foreign key, that it will always have a value and it must map to a reference entry, and in reference the curie is a required field.  referencefile also has an on delete cascade off of the reference_id from the reference table, so I think the value must exist.  Would be good to get confirmation from Valerio though, thanks for bringing it up :)

agr_literature_service/api/crud/referencefile_crud.py
```diff
@@ -0,0 +1,72 @@
...
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                            detail=f"Reference with curie {request.reference_curie} does not exist")
+    del request_dict["reference_curie"]
+    request_dict["reference_id"] = ref_obj.reference_id
+    mod_abbreviation = request_dict["mod_abbreviation"]
+    del request_dict["mod_abbreviation"]
+    new_ref_file_obj = ReferencefileModel(**request_dict)
+    db.add(new_ref_file_obj)
+    db.commit()
+    if mod_abbreviation is not None:
```
azurebrd, line 30: This isn't right, is it ?  We want to be able to create a referencefile with a null mod, which would create a referencefile_mod with a null mod, meaning it's PMC ?  Or for the PMC case would we require creating the referencefile first, then explicitly creating the referencefile_mod with a null mod ?   
Same issue with the referencefile_mod create.  
We should probably add a test for that.

azurebrd (commented):
Looks good !  I just had that question about how we should handle creating a referencefile for PMC, meaning the mod is `null`, and whether that's currently possible.  It might be that we're making things harder on ourselves by trying to handle null as data, instead of making null be null, and figuring out a way to get the mod table to hold non-mod values like PMC or PMID or whatever we might need.

sweng66:
@azurebrd Good point about null mod_id ! I totally forgot about it.

Thanks for the explanations for my questions. All make sense !!

azurebrd:
Thanks, @sweng66 !

tests/api/test_referencefile.py
```diff
@@ -0,0 +1,101 @@
...
+    print("***** Adding a test referencefile *****")
+    with TestClient(app) as client:
+        new_referencefile = {
+            "display_name": "Bob",
+            "reference_curie": test_reference.new_ref_curie,
+            "file_class": "main",
+            "file_publication_status": "final",
+            "file_extension": "pdf",
+            "pdf_type": "pdf",
+            "md5sum": "1234567890"
```
azurebrd, line 30: Might want another test with is_annotation true/false, since it's a different type of variable, and maybe another test with a null pdf_type to check that also works

sweng66:
__table_args__ = (
        Index('idx_md5sum', 'md5sum', unique=True),
        Index('idx_reference_id_display_name', 'reference_id', 'display_name', unique=True),
 )

Just realized this.

We need to change unique constraint a bit. Instead of setting reference_id + display_name unique, we have to set reference_id + display_name + file_extension unique,  or set display_name = file_name_with_suffix (eg, 1752-0509-1-35-1.gif) instead of "1752-0509-1-35-1". See an example PMC OA package below for PMID:17678529, PMCID:PMC1995223:

1752-0509-1-35-1.gif
1752-0509-1-35-1.jpg
1752-0509-1-35-2.gif
1752-0509-1-35-2.jpg
1752-0509-1-35-3.gif
1752-0509-1-35-3.jpg
1752-0509-1-35-4.gif
1752-0509-1-35-4.jpg
1752-0509-1-35.nxml
1752-0509-1-35.pdf

This is pretty common in PMC OA packages and other suppl files.

azurebrd:
That's a great point, @sweng66 , in my mind the file names had the extensions, but that's definitely not what we said.  Thanks for catching this.

sweng66:
In the google doc,  we have the following.

"Display_name - without extension"

So either change it to  

"Display_name - with extension"

or adding file_extension to the unique constraint

azurebrd:
Yeah, I think it makes sense to have the name without extension in the field, to properly tokenize it, and save space, and we should do what you said, @sweng66 , and change the constraint

sweng66:
OK sounds good! @azurebrd

valearna:
Thanks a lot @azurebrd and @sweng66. I forgot about PMC. I've modified the code to accept None as a mod_abbreviation. Now a referencefile_mod entry is always created when a new referencefile is added and the mod is set to None (PMC) if none is provided. I've added the tests you suggested @azurebrd and also additional tests to make sure adding a new referencefile_mod entry to an existing referencefile works as expected. I've also modified the unique constraint index in referencefile to include file_extension.

sweng66 (approved):
Looks great! @valearna

azurebrd (dismissed):
Thanks @valearna !

azurebrd (approved):

agr_literature_service/api/crud/referencefile_crud.py
valearna: fixed by allowing null values for PMC

tests/api/test_referencefile.py
valearna: added the suggested tests, thanks @azurebrd

PR #281, opened 2022-11-07
SCRUM-2270 fix(github_actions): upgraded checkout module from v2 to v3

valearna:


sweng66 (approved):
Looks good! Thanks @valearna

valearna:
@mluypaert do you have any recommendations on upgrading aws-actions? It seems that there's aws-actions/configure-aws-credentials@v1-node16 that uses node 16 instead of 12 to fix the deprecation issues, but I haven't found a solution for amazon-ecr-login

azurebrd (dismissed):
Thanks @valearna !

mluypaert:
> @mluypaert do you have any recommendations on upgrading aws-actions? It seems that there's aws-actions/configure-aws-credentials@v1-node16 that uses node 16 instead of 12 to fix the deprecation issues, but I haven't found a solution for amazon-ecr-login

Based on the [commit history](https://github.com/aws-actions/amazon-ecr-login/compare/v1.5.1...v1.5.2), it would seem that as of v1.5.2 `aws-actions/amazon-ecr-login` is supposed to use node.js v16, and the v1 release seems to point to the same commit as the latest release (v1.5.3), so should also be using this version. So I'm not sure why we are still seeing these warnings for this workflow, but I would assume they will go away automatically, so I would just keep an eye on that one.

`aws-actions/configure-aws-credentials@v1-node16` seems to be the current solution for that workflow, though this seems to be [a temporary fix](https://github.com/aws-actions/configure-aws-credentials/issues/489#issuecomment-1278145876), as an intrim solution between now and the future expected `v2`.

valearna:
Thanks a lot @mluypaert, I'll upgrade aws-actions/configure-aws-credentials to v1-node16 and keep aws-actions/amazon-ecr-login as is

azurebrd (approved):

PR #282, opened 2022-11-07
fix(referencefile): changed tag to referencefile and referencefile_mo…

valearna:
…d apis to be grouped under reference in the swagger ui


azurebrd (approved):
Thanks @valearna

PR #286, opened 2022-11-10
Scrum 2274 schema related fix

valearna:
removed duplicate fields from referencefile and referencefile_mod related schemas


azurebrd (approved):
Thansk @valearna !

PR #287, opened 2022-11-14
SCRUM-2148

pjhale:
Adding in authors facet... without filtering.  
Needs SCRUM-2148 in the agr_literature_ui to work properly.


valearna (approved):
thanks @pjhale!

PR #288, opened 2022-11-15
Scrum 2284 - added support for multiple referencefiles (file metadata) for the same md5sum

valearna:


sweng66:
op.drop_index('idx_md5sum', table_name='referencefile')

op.create_index('idx_md5sum', 'referencefile', ['md5sum'], unique=False)

so we need these two commands to drop this index. right? I thought first one will do that?

valearna:
Thanks for taking a look at this PR so wuickly @sweng66. I kept a non-unique index on md5sum since we want to search referencefiles (also) by md5sum and with an index the queries will be much faster

azurebrd (approved):

sweng66:
@valearna Oh I see. It is an index, not unique key. Makes sense

azurebrd (approved):

PR #289, opened 2022-11-15
Scrum 2258 lit md5sum

FB-ping:
refractory  of md5sum util


sweng66:
In save_database_md5sum function line 199, 

reference_id = primary_key

Assume the primary_key is an integer here. Maybe should fix the example reference_id to an integer?

sweng66:
In save_database_md5sum function line 222,

When you are updating the md5sum, I think we also need to update the date_updated

azurebrd (commented):
I'll defer to Shuai and Valerio.  It makes sense, but I don't think I follow the overall picture.

sweng66:
In save_database_md5sum function line 206,  this is not a big deal, but I think we can get rid of alias 'rmm' since we are only selecting data from one table (reference_mod_md5sum) and we are not using alias in "mod_id_sql" part.

sweng66:
Everything else looks good (in these two functions) Thanks. @FB-ping

sweng66:
Never mind. I think the date_updated will be updated automatically.

sweng66 (dismissed):

azurebrd:
> Never mind. I think the date_updated will be updated automatically.

Oh, would it ?  I thought that was only for AuditedObjects, and this wasn't.  (I don't know though)

sweng66:
@azurebrd https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-2258-lit-md5sum/agr_literature_service/api/models/reference_mod_md5sum_model.py#L55

Looks like it will update the date_updated automatically?

azurebrd:
> @azurebrd https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-2258-lit-md5sum/agr_literature_service/api/models/reference_mod_md5sum_model.py#L55
> 
> Looks like it will update the date_updated automatically?

Oh, cool, so that triggers on any existing row update ?  Nice, thanks

sweng66:
@azurebrd I think so.

agr_literature_service/lit_processing/data_ingest/dqm_ingest/utils/md5sum_utils.py
```diff
@@ -162,6 +161,133 @@ def save_s3_md5data(md5dict, mods):
...
+                    },
+                    "PMID": {
+                    "PMID:9241": "TEST5-6eac9538fafd9f73eff28dd0a28a2edf"
+                    }
+            }
+    """
+
+    md5dict = {}
+    mod_ids = {}
+    try:
```
valearna, line 259: do we need this try except? If so you should move the initialization of db_session outside it otherwise if there's an exception the session may be None

valearna: Also, you can rewrite line 259-264 with a one liner: `db_session = create_postgres_session(False)
    mod_ids = {mod_abbr_id["abbreviation"]: mod_abbr_id["mod_id"] for mod_abbr_id in db_session.execute("select abbreviation, mod_id from mod")}`

sweng66: I don't think we need try/except. They should work so I get rid of it. Good catch! Thanks @valearna

agr_literature_service/lit_processing/data_ingest/dqm_ingest/utils/md5sum_utils.py
```diff
@@ -162,6 +161,133 @@ def save_s3_md5data(md5dict, mods):
...
+                   "PMID": {
+                              "PMID:9241":      "6eac9538fafd9f73eff28dd0a28a2edf",
+                              12345:            "56782344a115c9ce612cab87869c1234"}
+    }
+    :param md5dict:
+    :param mods:
+    :return:
+    """
+    mod_ids = {}
+    try:
```
valearna, line 186: This could be turned into a oneliner, see comment below

agr_literature_service/lit_processing/data_ingest/dqm_ingest/utils/md5sum_utils.py
```diff
@@ -162,6 +161,133 @@ def save_s3_md5data(md5dict, mods):
...
+            logger.error("mod not in database yet:" + mod)
+            continue
+        mod_id_sql = "mod_id is null" if mod_id is None else f"mod_id = {mod_id}"
+        for primary_key in md5sums:
+            if primary_key.isnumeric():
+                reference_id = primary_key
+            else:
+                reference_results = db_session.execute(f"select reference_id from cross_reference r where r.curie = '{primary_key}' and is_obsolete='false' ")
+                refs = reference_results.fetchall()
+                if len(refs) == 0 :
```
valearna, line 212: extra whitespace after 0

agr_literature_service/lit_processing/data_ingest/dqm_ingest/utils/md5sum_utils.py
```diff
@@ -162,6 +161,133 @@ def save_s3_md5data(md5dict, mods):
...
+            md5sums = md5sum_results.fetchall()
+            if len(md5sums) == 0:
+                print("insert new md5sum: " + mod + " primary_key:" + str(primary_key) + ' ' + md5dict[mod][primary_key])
+                if mod_id is None:
+                    db_session.execute(f"insert into reference_mod_md5sum(reference_id, md5sum, date_updated) values ({reference_id}, '{md5dict[mod][primary_key]}', 'now()') ")
+                else:
+                    db_session.execute(f"insert into reference_mod_md5sum(reference_id, mod_id, md5sum, date_updated) values ({reference_id}, {mod_id}, '{md5dict[mod][primary_key]}', 'now()') ")
+                continue
+            md5sum = md5sums[0]["md5sum"]
+            if (md5sum == md5dict[mod][primary_key]):
```
valearna, line 225: you can remove the extra parentheses

agr_literature_service/lit_processing/data_ingest/dqm_ingest/utils/md5sum_utils.py
```diff
@@ -162,6 +161,133 @@ def save_s3_md5data(md5dict, mods):
...
+                continue
+            else:
+                print('need to update: ' + mod + '->' + str(primary_key) + ' old:' + md5sum + '->new:' + md5dict[mod][primary_key])
+                reference_mod_md5sum_id = md5sums[0]["reference_mod_md5sum_id"]
+                try:
```
valearna, line 230: what happens if we catch an exception here? Shouldn't we stop the execution?

sweng66 (approved):
everything looks great! Thanks @FB-ping @valearna @azurebrd

azurebrd:
Thanks @sweng66 !

PR #291, opened 2022-11-16
fix 1 bug for md5sum insert

FB-ping:


sweng66:
I think everything is working fine so no need this fix! @FB-ping @valearna

FB-ping:
@sweng66 @valearna Yes, without the change, it will fail to load new PMID , the SQL for null value insertion is wrong.

azurebrd:
@FB-ping Shuai tested loading the data after Valerio's updates, and says that it worked well.  Do you have an example that doesn't work after Valerio's changes ?

azurebrd:
The four of us talked, and everything's fine

PR #292, opened 2022-11-17
fixed to handle one md5sum to many reference_id mapping

sweng66:


azurebrd (approved):
Thanks @sweng66 .  I'm not sure I follow why it's happening, but it's changing the keys from some values to tuples of other values ?

sweng66:
Thanks for looking over the fixes. @azurebrd 

md5sum was an unique key in the referencefile table. now the combination of md5sum and reference_id is unique.

PR #293, opened 2022-11-18
Get dqm code ready for cron job

sweng66:


valearna (approved):
Everything looks good to me, thanks @sweng66

agr_literature_service/lit_processing/data_ingest/post_reference_to_db.py
```diff
@@ -66,6 +66,9 @@ def read_data_and_load_references(db_session, json_data, journal_to_resource_id,
     for entry in json_data:
 
         primaryId = set_primaryId(entry)
+        crossRef = db_session.query(CrossReferenceModel).filter_by(curie=primaryId).one_or_none()
```
azurebrd, line 69: I'm not sure what this is for, seems to be avoid entering something if it's already there ?  If so, should it be only if is_obsolete==false ?

azurebrd (commented):
Thanks @sweng66 and @valearna !  I just had that one comment about the xref filter for is_obsolete

agr_literature_service/lit_processing/data_ingest/post_reference_to_db.py
sweng66: Good catch! @azurebrd

sweng66: Fixed =) @azurebrd

azurebrd (approved):
Thanks @sweng66 !

PR #298, opened 2022-11-22
fix(referencefile): using faster loading strategy for referencefile a…

valearna:
…nd referencefile_mod sub-queries

### DIFF
--- a/agr_literature_service/api/crud/reference_utils.py
+++ b/agr_literature_service/api/crud/reference_utils.py
@@ -1,10 +1,10 @@
 import logging
 
 from sqlalchemy import or_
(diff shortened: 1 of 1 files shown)


valearna:
The current loading strategy for referencefile and referencefile_mod (lazy load) in the reference show api is very slow, and this is an issue for references with a lot of files (e.g., https://stage-literature-rest.alliancegenome.org/reference/535221). The new strategy reduces the loading time from ~ 1 min 45 sec to ~3 sec (measured running the api locally and connecting to rdsdev, which is slower than stage and dev)

azurebrd (approved):
Wow, such improvement, thanks @valearna !

PR #299, opened 2022-11-22
feat(referencefile): using standard s3 storage instead of glacier_ir …

valearna:
…when in dev mode


azurebrd (approved):
Great, thanks @valearna !

PR #304, opened 2022-12-02
adding MGI to the dqm loading pipeline

sweng66:
and fixing dqm loading script to ignore "bad" DOIs ("DOI:10.1042" and "DOI:10.1042/")


azurebrd (approved):
Thanks @sweng66 !

PR #305, opened 2022-12-05
cleaning up DOIs from MOD dqm files before comparing and loading and

sweng66:
fixing code to only use valid XREF IDs when comparing the data in ABC and in dqm files;
also cleaning up code for OLD & NEW md5sum checking for PubMed updating script


azurebrd (approved):
Looks good, thanks @sweng66 !   (I'm not sure I follow why things are happening, but overall it makes sense)

sweng66:
Just cleaning up some code and cleaning DOI IDs from dqm before checking and loading. Some DOIs from mods have spaces in them and some have leading root url in the DOIs so need to clean them up. Thanks @azurebrd

PR #307, opened 2022-12-06
Scrum 2165

azurebrd:


sweng66:
Looks great! Thanks @azurebrd 

usually I would add the following after line80
db_session.add(x)
I didn't know it would work without it.

sweng66 (approved):

azurebrd:
@sweng66 Oh.  I think it worked ?  I mean, I think I saw the db change afterward.  @valearna do you know if that should work ?

sweng66:
@azurebrd I believe you. I just didn't know that before.

azurebrd:
@sweng66 It's kind of late, I was just starting to doubt whether I'd set things backwards in the data (at first it was backward, because I was checking that the wrong thing was still on 3002), and I was checking the wrong thing :P  The logs make sense though.  Thanks :)

sweng66:
I see.

PR #308, opened 2022-12-06
fix(docker): fixed postgres installation on dev app

valearna:


sweng66 (approved):
Wow! only need one command now @valearna

azurebrd (commented):
Thanks @valearna !

PR #309, opened 2022-12-06
Scrum 2275 - new API to download tarball with additional files

valearna:


azurebrd (approved):
Thanks @valearna 
I wondered before why you needed a lambda function for the cleanup.  Did that not work, or do you think it's better to have an explicit function that is only used in one place ?  (Everything's fine, just wondering)

valearna:
thanks @azurebrd. Yes, the lambda function was not working so now I'm passing an existing function to do the cleanup

azurebrd:
@valearna gotcha, thanks

PR #311, opened 2022-12-06
a new script to identify main PDF files for newly downloaded PMC files

sweng66:
and adding it to the PMC package download pipeline

### DIFF
--- a/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/pubmed_identify_main_pdfs.py
+++ b/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/pubmed_identify_main_pdfs.py
@@ -1,0 +1,99 @@
+import logging
+import time
+import requests
+from agr_literature_service.lit_processing.utils.sqlalchemy_utils import \
+    create_postgres_session
+from agr_literature_service.api.models import CrossReferenceModel, ReferencefileModel
+
+logging.basicConfig(format='%(message)s')
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+rootUrl = "https://www.ncbi.nlm.nih.gov"
+pmcRootUrl = 'https://ftp.ncbi.nlm.nih.gov/pub/pmc/'
+
+infile = "data/pmc_oa_files_uploaded.txt"
+batch_size = 20
+
+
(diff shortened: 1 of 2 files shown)


agr_literature_service/lit_processing/data_ingest/pubmed_ingest/pubmed_identify_main_pdfs.py
```diff
@@ -0,0 +1,99 @@
...
+from agr_literature_service.api.models import CrossReferenceModel, ReferencefileModel
+
+logging.basicConfig(format='%(message)s')
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+rootUrl = "https://www.ncbi.nlm.nih.gov"
+pmcRootUrl = 'https://ftp.ncbi.nlm.nih.gov/pub/pmc/'
+
+infile = "data/pmc_oa_files_uploaded.txt"
```
azurebrd, line 15: Does this file need to be in the repo ?

sweng66: No, it is created by main PMC downloading script that calls the function in this script. This file will be cleaned up in the next run (main script is running via daily).

azurebrd (approved):
Thanks @sweng66 .  I don't know what the source data is like, so I can't tell if that makes sense, but I assume it works, since you and curators tested it.  

My only question is about infile = "data/pmc_oa_files_uploaded.txt"   whether it needs to be in the repo, or at least have a comment about what script would generate that file if running from a cronjob ?

agr_literature_service/lit_processing/data_ingest/pubmed_ingest/pubmed_identify_main_pdfs.py
azurebrd: Gotcha, sounds good.

sweng66:
Good suggestion! Just added comments to explain the source of the input file. Thanks @azurebrd

azurebrd (approved):

valearna (approved):

sweng66:
Thanks @azurebrd @azurebrd

PR #315, opened 2022-12-08
Scrum-1355 Highlighting

pjhale:
Adds API code for highlighting and only returns a highlight section if the request contains a query.


azurebrd (approved):
Thanks @pjhale

valearna (approved):
thanks @pjhale !

PR #316, opened 2022-12-09
fixed to not do anything to the database if there is any XREF ID conf…

sweng66:
…lict


azurebrd (commented):
So if there's an xref conflict, it never gets a chance to aggregate_mod_specific_fields_only nor aggregate_mod_biblio_all nor write reports about PMIDs and DOIs ?  
I didn't catch that Ceri and Kimberly / curators wanted that, but if they do, then okay.

sweng66:
Good question! I have chatted with Ceri about this this PM. It will just not update any data for the given record in the database, but it will report any XREF ID conflict and will write log entries for it. @azurebrd

azurebrd (approved):
Ah, okay, so long as @cerivs understands that other data for the paper won't get updated if there's an xref conflict, as opposed to just not updating the xref, then cool.  Thanks @sweng66

PR #318, opened 2022-12-12
instructions on setting up a small data subset of 2000 references to …

ianlongden:
…enable elastic search setup locally for testing and development


sweng66 (approved):
Thanks for the notes @ianlongden

PR #320, opened 2023-01-12
adding sql code to retrieve the pmids that do not have a PMC package …

sweng66:
…downloaded for all papers in corpus with a PMCID


azurebrd (dismissed):
Thanks @sweng66 !

sweng66:
Thanks so much! @azurebrd

azurebrd (approved):

PR #322, opened 2023-01-13
Scrum 2463 - remove dependency on agr_base_linux_env from ABC docker images

valearna:


azurebrd (approved):
Thanks so much, @valearna !

PR #323, opened 2023-01-13
Scrum 2463 adding missing packages to xml_processing image

valearna:


valearna:
I forgot to add some packages to the xml_processing image. They are needed to install some python requirements (e.g., psycopg2-binary)

ianlongden (approved):

PR #325, opened 2023-01-17
script to clean up RESOURCE data

sweng66:


agr_literature_service/lit_processing/oneoff_scripts/cleanup_resource_data.py
```diff
@@ -0,0 +1,157 @@
...
+    for x in rows:
+        resource_ids_with_XREF.add(x[0])
+
+    row_count = 0
+    for duplicate_resource_id in duplicate_resource_id_to_correct_resource_id:
+        if duplicate_resource_id in resource_ids_with_XREF:
+            logger.info("duplicate_resource_id: " + str(duplicate_resource_id) + " is still in CROSS_REFERENCE table so keep it")
+        else:
+            row_count += 1
+            x = db_session.query(ResourceModel).filter_by(resource_id=duplicate_resource_id).one_or_none()
```
azurebrd, line 142: @sweng66 : I think @ianlongden thinks we should also delete from resource_version here, based on the duplicate_resource_id

sweng66: I just have a second thought on this.. 

Since this is a such small set comparing with other data in the history tables, I am reluctant to remove them at this moment in its version table - just in case we want to look at some of these data later. 

Of course I can remove these data if you all think we should @azurebrd @ianlongden

ianlongden: The version table should have a deleted status at the end I think so we can do the deletion of the version table later on, when you are happy all went okay. In theory at least :-)

sweng66: that sounds good @ianlongden

sweng66: We have A LOT of RANDOM data in the _version tables that we should remove

ianlongden: It might be safer if not deleting the version table to dump out the resource_ids of those deleted to a log file as you go along. Just to be extra "safe" that we can clean up the version table later.

ianlongden: Actually we should be able to do a left join on the version table later to clean up properly anyway.

ianlongden: More precisely from the resource_version table left join to the resource table and delete where no resource found (NULL).

azurebrd: I'll defer to you both on what's best.
If 80% of our resource data is bad, removing it from the version table would make sense if it's a lot of data needlessly bloating the database, but since resource objects are not that big, and there's not nearly as many as references, maybe it's just fine to leave it for future cleanup.

agr_literature_service/lit_processing/oneoff_scripts/cleanup_resource_data.py
```diff
@@ -0,0 +1,157 @@
...
+                              "WHERE curie_prefix = 'NLM' and is_obsolete = False").fetchall()
+    for x in rows:
+        resource_id_to_nlm[x[0]] = x[1]
+
+    duplicate_resource_id_to_correct_resource_id = get_duplicate_resources(db_session,
+                                                                           resource_id_to_nlm)
+
+    logger.info("Updating resource_ids in EDITOR table...")
+
+    # update_editor_table(db_session, duplicate_resource_id_to_correct_resource_id)
```
azurebrd, line 28: I see the code exists for update_editor_table and update_reference_table, I'm not sure if it's commented out because it shouldn't be run, or to treat the review as if it's going to be run but it was commented out in testing.
I would think you want to delete all the editors that exist in duplicate resources, but that you'd want to just remove the reference-resource connection for duplicate resources.
I think right now it's doing the same thing for both, mapping editors and references to the new resource, but I don't think we need to.  An editor should only exist for a given resource, not multiple resources (as opposed to a person that could be connected to both), so I think it can be safely deleted.  For reference-resource connections, I would think that the reference is already connected to the correct resource and doesn't need to be reconnected, but I haven't looked at the data, but I guess it'd be good to check whether it's making a duplicate reference-resource connection (I don't even know if that's possible, it might not be, in which case it's probably fine).

azurebrd (commented):
I think everything's fine, if that's how you'd like it, @sweng66 .  I made another single comment about the commented out editor/reference connections, but since they're commented out, that's fine.  I think it'd be better to remove any editors related to the bad resources, and just remove the reference-resource connections for the bad resources.  If some references end up without resources, I would think that a future run of an existing script that gets their data would connect them to the correct resource, but maybe that's not true.
Deferring to you and and Ian about the _version tables.

agr_literature_service/lit_processing/oneoff_scripts/cleanup_resource_data.py
sweng66: @azurebrd Yes, just feel more comfortable to keep them for now so we can delete these extra rows when everything is fine.

azurebrd: cool, sounds good

agr_literature_service/lit_processing/oneoff_scripts/cleanup_resource_data.py
sweng66: @azurebrd oops good catch!! lol 

Just commented out the two update statements.

sweng66:
@azurebrd Thanks for the suggestion!

So what I am going to do is to fix the script to remove the editor rows (linked with the duplicate resource_ids) (like you have suggested) and leave the code to fix the 'bad' reference/resource connection part as is (so we can fix the bad connections for any non-pubmed papers from dqm loading, any issues with pubmed papers (I don't think it will have) will be fixed by weekly pubmed update script). This way, we need to deal with data from dqm specifically- just keep cronjob running as is. For _version tables, we can delete these and other unwanted data at some point at the same time in the future. 

Does this sound ok?

sweng66:
If you think it is better to just unlink the reference and bad resource (instead of fixing them), I can do that as well.

sweng66:
Just fixed the program to do the following based on the suggestions from @azurebrd @ianlongden. Thanks both for going over the code!!

1. remove the EDITOR rows that are associated with a duplicate resource entry
2. set resource_id to None if the Reference entry is linked to a duplicate resource row
3. remove the duplicate RESOURCE rows from RESOURCE table
4. remove the duplicate RESOURCE rows from RESOURCE_VERSION table

azurebrd:

sweng66:
fixed the script again to remap the duplicate resource_ids to correct ones in the REFERENCE table instead of setting them to NULL to avoid going through every paper in the database during DQM reference loading and PubMed update.

agr_literature_service/lit_processing/oneoff_scripts/cleanup_resource_data.py
sweng66: This is solved

agr_literature_service/lit_processing/oneoff_scripts/cleanup_resource_data.py
sweng66: This is solved.

sweng66:
@azurebrd Many thanks for spending time going through all the details.

azurebrd:
Thanks for all the work, @sweng66 !

sweng66:
@valearna @ianlongden Any concerns for me to merge the code to main?

ianlongden (approved):

PR #327, opened 2023-01-19
Add sequence data else you cannot add new entries.

ianlongden:
For some reason sequences are not updated when data is added to the subset database. So if you try to add a new one it try to use an already existing primary key. This fix just sets them all (better safe than sorry) to a fairly high number.

### DIFF
--- a/agr_literature_service/lit_processing/oneoff_scripts/make_subset_db/make_subset.py
+++ b/agr_literature_service/lit_processing/oneoff_scripts/make_subset_db/make_subset.py
@@ -118,8 +118,9 @@ def start():
         "workflow_tag_reference_workflow_tag_id_seq"
     ]
     for seq in seq_list:
-        com  = f"ALTER SEQUENCE {seq} RESTART WITH {num_of_ref+1000}"
+        com = f"ALTER SEQUENCE {seq} RESTART WITH {num_of_ref+1000}"
         db_subset_session.execute(com)
     db_subset_session.commit()
+
 
 start()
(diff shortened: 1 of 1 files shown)


azurebrd:
What's the subset database ?

sweng66:
@ianlongden Did you mean you just load a subset of data into your local database? Does the increase seq by 1000 mean that the max rows for each table is 1000?

ianlongden:
> @ianlongden Did you mean you just load a subset of data into your local database? This is probably a dumb question..does the increase seq by 1000 mean that the max rows for each table is 1000?

It means the next one added will have XXX_id of 3001  (num of refs + 1000). This is for the subset database for testing search easily. (num_of_refs is 2000 by default.) So if I add a new reference it will have reference_id of 3001, previously it was 1 which was already used.

ianlongden:
> @ianlongden Did you mean you just load a subset of data into your local database? This is probably a dumb question..does the increase seq by 1000 mean that the max rows for each table is 1000?

The script the is in copy the first 2000 references and its linked data into a new database which is called literature_subset by default.

sweng66:
I am confused. If previously it was 1, then adding 1000, wouldn't it be 1001?
Anyway, if it is working for you, then it is great. Thanks

ianlongden:
> I am confused. If previously it was 1, then adding 1000, wouldn't it be 1001?
> Anyway, if it is working for you, then it is great. Thanks

We are adding 2000 references to the new subset database so I add 1000 to that to make the sequence start at 3001. 
ALTER SEQUENCE {seq} RESTART WITH {num_of_ref+1000}"
seq is from a list of all the XXX_seq tables.
The sequence is used on auto increment for several primary kets like reference_id in references.
Hopefully that makes things clearer. If you are confused by anything you are reviewing then it is the programmers fault for not explaining properly. I got lazy as its a one off script.

sweng66:
Lollll just saw num_of_ref + 1000.  That's great! @ianlongden 

Thanks

sweng66 (approved):

PR #331, opened 2023-01-24
fix(search): inserting "should" query under main "must" query

valearna:
fixes bug in search: when searching on "all text fields" facets are not working correctly


sweng66 (approved):
I was looking at this section earlier. Thanks @valearna

PR #333, opened 2023-01-27
SCRUM-1302

ianlongden:


agr_literature_service/api/crud/search_crud.py
```diff
@@ -10,13 +12,90 @@
...
+    if date_published:
+        try:
+            start = date_str_to_micro_seconds(date_published[0], True)
+            end = date_str_to_micro_seconds(date_published[1], False)
+        except Exception as e:
+            logger.error(f"Exception in conversion {e} for start={date_published[0]} and end={date_published[1]}")
+        logger.debug(f"Search date_published: start={start}, end={end}")
+        es_body["query"]["bool"]["filter"]["bool"]["must"].append(
+            {
+                "bool": {
```
pjhale, line 69: I think this could be a problem here... it might need to have this [should] clause appended at the same level as the [must] clause.  Although I'm not sure why [should] is needed in a case of filtering.  Wouldn't this always be must?

agr_literature_service/api/crud/search_crud.py
```diff
@@ -10,13 +12,90 @@
...
+    # as elastic search does a tr comparison and if the end date has the time
+    # element stil in it fails even if the date bits match as the string is longer.
+    if "must" not in es_body["query"]["bool"]["filter"]["bool"]:
+        es_body["query"]["bool"]["filter"]["bool"]["must"] = []
+    if date_pubmed_modified:
+        es_body["query"]["bool"]["filter"]["bool"]["must"].append(
+            {
+                "range": {
+                    "date_last_modified_in_pubmed": {
+                        "gte": date_pubmed_modified[0].split('T')[0],
```
pjhale, line 45: Is this necessary?  I didn't think these dates had a time component?

agr_literature_service/api/crud/search_crud.py
ianlongden: This bit just needs one of the ranges to be true, so a should (or) is needed I think. Either the start or end date needs to in-between the values given.

agr_literature_service/api/crud/search_crud.py
ianlongden: The values in the database do not have time component but the ui returns a string with the time components which messes things up in the search as one string is longer/bigger than the other if the dates match.

agr_literature_service/api/crud/search_crud.py
```diff
@@ -10,13 +12,90 @@
...
+                            "range": {
+                                "date_published_end": {
+                                    "gte": start,
+                                    "lte": end
+                                }
+                            }
+                        },
+                        {
+                            "range": {
+                                "date_published_start": {
```
pjhale, line 81: I don't know if we need to take into account both date published start and date published end.  Even if we do I think these would need to be in separate clauses.  I would check with Ceri on this but for the sorting story we are only concerned with date published start.

ianlongden: For filtering we need to check both I think. Actually I may need another range too.
 There are 3 cases where we want to find papers
1) pub_start is between dates selected,
2) pub_end is between dates selected
3) pub_start > start date selected AND pub_end < end date selected.

agr_literature_service/api/crud/search_crud.py
```diff
@@ -10,13 +12,110 @@
...
+                "range": {
+                    "date_arrived_in_pubmed": {
+                        "gte": date_pubmed_arrive[0].split('T')[0],
+                        "lte": date_pubmed_arrive[1].split('T')[0]
+                    }
+                }
+            })
+    if date_published:
+        try:
+            start = date_str_to_micro_seconds(date_published[0], True)
```
valearna, line 62: Why do we need to treat the date_published dates differently? They configuration in the es mappings is the same as the other pubmed dates so they should work in the same way

valearna (approved):
Thanks @ianlongden all the rest looks good!

agr_literature_service/api/crud/search_crud.py
ianlongden: I will merge this now and review just using a date later.  I probably need to look at other dates while I am there anyway and want to get this in so I can get other stuff done and go back to this. Thanks.

PR #339, opened 2023-02-06
Scrum 2474 - api endpoints to retrieve and paginate topic and entity tags

valearna:
- New api endpoint to retrieve list of tags (or get the total count) - took it out from the main reference endpoint
- endpoint to get curie -> name mappings from a-team - removed from UI


azurebrd (approved):
Thanks @valearna !

sweng66 (approved):
Code looks good although I can follow everything. Thanks @valearna

PR #340, opened 2023-02-06
Set time zone to reflect data.

ianlongden:


valearna (approved):
I don't really understand what's going on with these dates but I trust you @ianlongden

ianlongden:
I do not understand either but when it's on stage we can see if that worked. It does not work currently so cannot be any worse.
Tests and stage act differently, I am presuming something somewhere is in the wrong timezone but no idea where and we have both looked!!!

PR #343, opened 2023-02-08
feat(restart_scripts): debezium scripts in makefile now stop dbz before starting services + longer sleep

valearna:
…re starting it + larger dbz sleep


sweng66 (approved):
Looks good! Thanks @valearna

valearna:
thanks @sweng66 !

PR #345, opened 2023-02-08
Scrum 2473 - added parameters to sort by entity type to topic and entity tags endpoint

valearna:
- sort by works only on entity type for now
- asc or desc through a second parameter


azurebrd (approved):
Thanks @valearna !

PR #347, opened 2023-02-13
Scrum 1980

valearna:
upload tar gz with multiple files

- files with extension tgz or gz are now decompressed on the fly and all files in them are uploaded singularly
- in case one or more single file uploads fail, an overall error message is returned with the details of what went wrong for each file
- If a single upload fails, the program continues with the remaining uploads


azurebrd (approved):
Thanks @valearna !

sweng66 (approved):
Looks great! Thanks @valearna

PR #349, opened 2023-02-13
Scrum 1980 - added support for .tar.gz files

valearna:


sweng66 (approved):
Looks good! Thanks @valearna

PR #351, opened 2023-02-21
Scrum-2489 Adding Parital Match Support

pjhale:
Adds support for partial matches and changes query type to simple_query_string.


valearna (approved):
Thanks @pjhale, everything is so much better with simple query!

PR #353, opened 2023-02-23
fixed the partial index for REF XREF IDs to exclude CGC IDs

sweng66:


valearna (approved):

sweng66:
Thanks @azurebrd @valearna

PR #354, opened 2023-02-24
SCRUM-2369 Adding Citation

pjhale:
Just adding citation to indexed fields and adding to all search.  

No UI changes.


valearna (approved):
thanks @pjhale !

PR #358, opened 2023-02-28
Scrum 2228 - new api endpoint to get list of reference files

valearna:
reference files now excluded from main reference api


azurebrd (approved):
Thanks @valearna !

PR #360, opened 2023-02-28
Scrum 2634 - removed default relationship loading strategies from reference model

valearna:
- added them as parameters to the get_reference method in the API
- added more options to queries in lit_processing utils


valearna:
@sweng66  I added more options to the reference query in lit_processing utils. Removing the default loading options makes queries to the base reference model faster when no additional info is needed. We can talk about this before merging this PR

azurebrd (approved):
Thanks @valearna .  Those `options` are neat.  I'll ask you again sometime why in one case it's subqueryload, and in others it's joinedload.  I didn't look so carefully for logic changes to what looked like a lot of formatting changes.

valearna:
I'm using subqueryload because it's the fastest option for those tables. When reading a single reference and loading all its referencefiles, a join takes a lot of time because the two tables are huge. On the other hand, subqueryload queries the reference first and then loads all the related referencefiles with a second query and this is much faster when reading a small number of references at the same time (in most of our api cases it's only one).

azurebrd:
Cool, thanks !

PR #361, opened 2023-03-01
add code to fix the date_published_start and _end from DateTime to String

sweng66:


azurebrd (commented):
Thanks @sweng66 .  I didn't really understand it, I thought we didn't need to make this change, so I'll defer to @valearna

sweng66:
@azurebrd Oh I thought we have some search issue related to the type. 

And we do not need to update the Search code to work with this type change based on Valerio (during standup).

I may miss something.  If I do, we need to recreate DEV from PROD since the alembic change is in DEV now.

Thanks Juancarlos for checking this out.   @valearna @pjhale do we need to change the type for date_published_start and date_published_end (from datetime to string)?

valearna (approved):

PR #362, opened 2023-03-01
fix(reference): changed field types for date_published start and end …

valearna:
…in reference schema


sweng66 (approved):
@valearna Good catch! Thanks

sweng66 (approved):
Great! Thanks @valearna

PR #363, opened 2023-03-02
one-off script to remove the trailing timestamp for date_published_start and date_published_end

sweng66:
also fixed the dqm loading and reference loading module to handle these two dates properly.


valearna (approved):
Thanks @sweng66 !

PR #365, opened 2023-03-03
Scrum 2586 fix htmlentity subset

FB-ping:
script to update author table first_name, name, last_name, affiliation html entity


cerivs:
The story was to only deal with the author representation.  I don't think drosophila is ever an author.

FB-ping:
@cerivs yes,  for those cases I show , only the first case exist in author table ( I make up those for author to generalize consequence), all other cases originate from reference abstract and title, which is not covered for this ticket.
BEFORE                       ----------  AFTER
Nemčovičová ----------  Nemčovičová
in &lt;i&gt;Drosophila&lt;/i&gt;   ----------    in <i>Drosophila</i>
α-mannosidase   ----------   α-mannosidase
tcf21<sup>+</sup>   ----------  tcf21<sup>+</sup>
<a href='/locus/S000004219'>Cdc42</a>   ----------    <a href='/locus/S000004219'>Cdc42</a>

sweng66 (commented):
Thanks @FB-pin ! The code looks good (I skip'ed the test code - since we do not run unit test against oneoff scripts). I have two questions..

1. I thought we just need to fix the author at this time, am I missing anything?
2. If we update the author / title / year + volume + issue / journal for a paper, we need to make sure the corresponding citation column is also up to date. When you ran your script, did you check the citation column? since the pubmed update script will update the citation column only if any of these (author/title/year/journal+etc) got changed. So if any of these citation related values got fixed in the database, then pubmed update script will miss the change. Does this make sense? or am I missing something? Ignore this question if the citation is already updated according

One minor thing:  we usually use "logger" instead of "print" to output info, but this is one off script so it is ok.

FB-ping:
@sweng66  1. yes, we only update author for this ticket, here there is function for reference, but in the main, we only call the update_author_entity(), so will NOT update anything in reference.  2. this (also the xml_to_json script) only update those columns of author: first_name, last_name, name, affiliation,  NOT touch anything of reference self.   citation got updated when we update author.name, I think this was triggered by some trigger function ?

sweng66:
@FB-ping If the author.name change will trigger the citation update, then this is good to go. Thanks Pingle!

I was wondering where the code is for triggering the citation update.

valearna (approved):
Looks good to me, thanks @FB-ping

PR #368, opened 2023-03-07
Scrum 2611 - topic and entity tag api now using 'page' and 'page_size' instead of 'offset' and 'limit'

valearna:
page starts from 1 and is set to 1 if the provided value is < 1


sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):
Thanks @valearna !

valearna:
thanks @azurebrd and @sweng66

PR #371, opened 2023-03-10
fixed reference api to return data even with an invalid XREF ID

sweng66:
and also return a new field 'invalid_cross_reference_ids' (a list of invalid id(s) or an empty list)


azurebrd (approved):
Great, thanks @sweng66 .
Did we ever find out why the API was failing when there were curies with 2 colons ?

sweng66:
Thanks @azurebrd for checking this out.

Yes, we have following check in the current cross_reference_schemas.py:
 
  if v.count(":") != 1 and not v.startswith("DOI:"):
         raise ValueError('must contain a single colon')

it will threw an error if a non DOI curie has more than one colon or has no colon.

azurebrd:
@sweng66 Ah, gotcha, didn't realize that was a thing.  Thanks !

sweng66:
@azurebrd Great!

PR #372, opened 2023-03-13
SCRUM-2699

ianlongden:
open access removed from reference and added to the copyright. Had to remove open_access in the schema files and tests.
At some point we should add copyright_license to these too.


sweng66 (approved):
@ianlongden Looks great to me. Thanks!

PR #374, opened 2023-03-14
fixing reference and referencefile APIs to deal with reference open access info

sweng66:


azurebrd (approved):
Thanks @sweng66.  I probably didn't understand everything, but it makes sense

sweng66:
@azurebrd @valearna Thanks both for checking out code. 
Yes to your question @valearna

PR #375, opened 2023-03-14
fix(open_access): fixed query to check if reference is open access with new table field

valearna:


sweng66 (approved):
Looks great! Thanks @valearna

PR #378, opened 2023-03-16
Scrum 2480 - bulk upload script for WB reference files

valearna:
Added a new docker image with a shell script to upload files to the API from a local directory with reference files organized on Daniel's local folder structure.

Modified referencefile API to upload files to accept cross references as for now WB needs to upload using WBPaperIDs, but the script uses AGR curies if the local reference files and directories have 10 digits.


azurebrd (approved):
That's great @valearna !
Does the upload_files.sh need to check that a MOD is entered else fail, or will it fail if the MOD is wrong/missing when it posts to the API ?  I can't tell where the MOD is the upload URL, but maybe it's not needed somehow ?  (I would think it's needed to give that MOD permission to see the files it uploaded)
I can't tell where BASH_REMATCH comes from, but I'm not too familiar with .sh files, I assume it's from the regex, but don't know how that variable comes into existence.

valearna:
Thanks @azurebrd! The script works even if you don't pass any MOD or a MOD abbreviation that does not exist, but the API call fails. I'm not recording the returned values into a log file, but I should at least print success and failure messages - I'll add this.
BASH_REMATCH is a bash specific variable that contains regex matches for groups. It is populated after a regex is evaluated using the =~ sign.
The way I tested the script to make sure that the files are parsed in the right way was to connect it to a wrong api endpoint so that all files are processed and the metadata is printed on stdout but the api calls fail. Maybe we can try to run it together and also discuss the output format for logging successes and failures?

azurebrd:
Sounds good, thanks @valearna !

azurebrd (approved):

PR #379, opened 2023-03-18
Scrum 2716 fixed loading strategies for reference query in script to subset db

valearna:
- Added individual loading strategies for each related table that needs to be loaded in the subset
- Using subqueryload makes the query much faster than joins
- Added separate queries for additional data that needs preloading


azurebrd (approved):
Thanks @valearna .  I don't recall what this script is for, but the subqueryload makes sense

PR #381, opened 2023-03-21
Scrum 2694

ianlongden:


valearna (dismissed):
Wow, this looks amazing @ianlongden!

sweng66 (commented):
Wow! you did so much! So many files got updated. Majority of the changes make sense and look good to me! Thanks @ianlongden 

I have one question about line 339 in post_reference_to_db.py.

https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-2694/agr_literature_service/lit_processing/data_ingest/post_reference_to_db.py#L339

You added a line to commit the new paper into 'reference' table. Originally, I purposely did not commit the data into 'reference' table. Instead, commit data after whole record got added into the database (not just reference table, all reference related tables, eg author, cross_reference etc) so we don't end up with incomplete reference record in ABC.  Any reason to do that (you may have mentioned this earlier)? I saw you expire the object then re-query the database. You have to do this to make trigger to create the citation, is that right?  wait.. but we still need authors etc for generating the citation. Confused.. Am I missing anything?

Anyway, great work on this new tricky trigger business!

valearna (approved):
Looks ready to be merged

PR #384, opened 2023-04-07
adding code to delete temp pdf files when a final main PDF is uploaded

sweng66:
by the same mod or downloaded from PubMed Central.
also fixing the 'resolve_displayname_conflict' function to append an increasing number to the base file name until a unique file name is found.

### DIFF
--- a/agr_literature_service/api/crud/referencefile_crud.py
+++ b/agr_literature_service/api/crud/referencefile_crud.py
@@ -160,32 +160,13 @@ def cleanup_temp_file(db: Session, ref_curie: str):  # pragma: no cover
                 ReferencefileModel.file_publication_status).all()
 
         if len(reffiles) >= 2:
-            modsWithFinal = []
-            pmcFinalPDF = False
-            for x in reffiles:
-                if x.file_publication_status == 'final':
-                    for m in x.referencefile_mods:
-                        if m.mod_id:
-                            if m.mod_id not in modsWithFinal:
-                                modsWithFinal.append(m.mod_id)
-                        else:
-                            pmcFinalPDF = True
-                if x.file_publication_status == 'temp':
-                    toDelete = False
-                    if pmcFinalPDF is True:
-                        toDelete = True
-                    else:
-                        for m in x.referencefile_mods:
-                            if m.mod_id in modsWithFinal:
-                                toDelete = True
-                    if toDelete is True:
-                        try:
-                            db.delete(x)
-                            remove_file_from_s3(x.md5sum)
-                            db.commit()
-                        except Exception as e:
-                            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-                                                detail=f"An error occurred when deleting temp pdf file. {e}")
+            mod_ids_with_final = {mod.mod_id for reffile in reffiles for mod in reffile.referencefile_mods if
+                                  reffile.file_publication_status == 'final'}
+            for reffile in reffiles:
+                if reffile.file_publication_status == 'temp':
+                    if None in mod_ids_with_final or all([mod.mod_id in mod_ids_with_final for mod in
+                                                          reffile.referencefile_mods]):
+                        destroy(db, reffile.referencefile_id)
 
 
 def file_upload_single(db: Session, metadata: dict, file: UploadFile):  # pragma: no cover
--- a/agr_literature_service/api/crud/referencefile_crud.py
+++ b/agr_literature_service/api/crud/referencefile_crud.py
@@ -111,6 +111,14 @@ def file_paths_in_dir(directory):
 
 
 def file_upload(db: Session, metadata: dict, file: UploadFile):  # pragma: no cover
+    if not metadata["reference_curie"].startswith("AGRKB:101"):
+        ref_curie_res = db.query(ReferenceModel.curie).filter(
+            ReferenceModel.cross_reference.any(CrossReferenceModel.curie == metadata["reference_curie"])).one_or_none()
+        if metadata["reference_curie"] is None:
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                                detail="The specified curie is not in the standard Alliance format and no cross "
+                                       "references match the specified value.")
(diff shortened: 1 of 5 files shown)


agr_literature_service/api/crud/referencefile_crud.py
```diff
@@ -141,6 +141,48 @@ def file_upload(db: Session, metadata: dict, file: UploadFile):  # pragma: no co
...
 
+    cleanup_temp_file(db, metadata["reference_curie"])
+
+
+def cleanup_temp_file(db: Session, ref_curie: str):  # pragma: no cover
+    if ref_curie is None:
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                            detail="The specified curie is not in the standard Alliance format and no cross "
+                            "references match the specified value.")
+    ref = db.query(ReferenceModel).filter_by(curie=ref_curie).one_or_none()
```
valearna, line 152: the curie passed in the metadata is either a reference curie or a xref, so if ref is None we should try to read the ref through the xref. We are already doing this in the function `file_upload_single` see lines 189-196. Maybe we can move this code at the beginning of the main function (`file_upload`) and then pass the reference object directly to `file_upload_single` and `cleanup_temp_file`.

sweng66: I see. Good catch! I will move code around. @valearna

agr_literature_service/api/crud/referencefile_crud.py
```diff
@@ -141,6 +141,48 @@ def file_upload(db: Session, metadata: dict, file: UploadFile):  # pragma: no co
...
+                                modsWithFinal.append(m.mod_id)
+                        else:
+                            pmcFinalPDF = True
+                if x.file_publication_status == 'temp':
+                    toDelete = False
+                    if pmcFinalPDF is True:
+                        toDelete = True
+                    else:
+                        for m in x.referencefile_mods:
+                            if m.mod_id in modsWithFinal:
```
valearna, line 175: This code is marking the file for deletion if there's at least one mod with a final file, but the final file might not be accessible by the other mods. We should delete only when all mods have access to a final pdf, or delete only the connection with the mod (maybe using the `destroy` function in this module?)

valearna: I'm thinking about something like this:

```reffiles = db.query(ReferencefileModel).filter_by(
            reference_id=ref.reference_id, file_class='main', file_extension='pdf').order_by(
                ReferencefileModel.file_publication_status).all()

        if len(reffiles) >= 2:
            mod_ids_with_final = {mod.mod_id for reffile in reffiles for mod in reffile.referencefile_mods if
                                  reffile.file_publication_status == 'final'}
            for reffile in reffiles:
                if reffile.file_publication_status == 'temp':
                    if None in mod_ids_with_final or all([mod.mod_id in mod_ids_with_final for mod in
                                                          reffile.referencefile_mods]):
                        destroy(db, reffile.referencefile_id)```

File: agr_literature_service/lit_processing/data_ingest/pubmed_ingest/load_pmc_metadata.py [hunk 3]
```diff
@@ -80,11 +82,28 @@ def load_ref_file_metadata_into_db():  # pragma: no cover
...
+    # base name to form the new display name.
+    #
+    # If the (reference_id, file_name_with_suffix) tuple is already present in the
+    # "ref_file_uniq_filename_set", we increment the counter, update the display name,
+    # and try again. We repeat this process until a unique file name is found.
 
     file_extension = file_name_with_suffix.split(".")[-1].lower()
-    display_name = file_name_with_suffix.replace("." + file_extension, "") + "_1"
-    return display_name + "." + file_extension
+    base_name = file_name_with_suffix.replace("." + file_extension, "")
```
valearna, line 98: This is the code I added to the function `file_upload_single` in the referencefile_crud:

```
original_name = metadata["display_name"]
        for counter in count(start=1, step=1):
            ref_file_same_name = db.query(ReferencefileModel).filter(
                and_(
                    ReferencefileModel.display_name == metadata["display_name"],
                    ReferencefileModel.file_extension == metadata["file_extension"],
                    ReferencefileModel.reference.has(ReferenceModel.curie == metadata["reference_curie"])
                )
            ).one_or_none()
            if ref_file_same_name is None:
                break
            metadata["display_name"] = f"{original_name}_{counter}"
```

sweng66: Wow. this looks great. Thanks @valearna !! 

In the PMC script, it needs to check the reference_id/display name pair for a long list of files so saving data in memory instead of checking one by one seems more efficient.

Anyway, I will leave the code as is.

agr_literature_service/api/crud/referencefile_crud.py
valearna: oh, and the same file (same md5sum) may be shared by other references, so we must use the `destroy` function

sweng66: Wow! it makes sense and the code is so much cleaner. Thanks ! @valearna 
I have implemented it. 

At first, I believed there would be no problem with the temp file uploaded by a mod being replaced by the final file provided by the same mod, but I guess if the paper is a shared one it would be a problem. wait...

sweng66: I need to think some more, but your implement should be safer.

agr_literature_service/api/crud/referencefile_crud.py
sweng66: Implemented.

agr_literature_service/api/crud/referencefile_crud.py
sweng66: "oh, and the same file (same md5sum) may be shared by other references, so we must use the destroy function"

Oh yes. I missed different papers can share files

valearna (approved):
Everything looks good now, thanks a lot @sweng66!

sweng66:
Many thanks to go over everything !!! Thanks @valearna

PR #385, opened 2023-04-10
Scrum 1971 - bulk file uploader

valearna:
Tested the bulk uploader on the whole WB corpus. 

Changes to the API:

- Fixed bug in code to read reference from cross reference when curie provided in upload metadata does not start with "AGRKB:101"
-  Overwrite "file_publication_status" and set it to "temp" when a temp file is uploaded and the same file (same md5sum) is already in the system but it's "final". Keep all the other metadata fields as they are in the db
- If the uploaded file has the same display name as a file already in the system associated to the specified reference but the existing file has different content (different md5sum), append an underscore followed by an incremental number to the display name of the uploaded file before storing it in the db, until there are no other files with the same name.

Changes to the file uploader:

- Added quotes to cope with whitespaces in file names
- Added function to generate a new okta token when the one in use expires


sweng66 (approved):
Looks good! Thanks @valearna

PR #387, opened 2023-04-11
fix(citation): added citation_id and citation_id_mod to reference_version

valearna:
changes to reference table were failing without these columns


sweng66 (approved):
Thanks @valearna

azurebrd (approved):

PR #392, opened 2023-04-22
fix(file_uploder): read supplements in subdirs

valearna:


sweng66 (approved):
Looks good! Thanks @valearna

PR #393, opened 2023-04-22
fix(file_uploader): allowing upload of files with $ in filename

valearna:


azurebrd (approved):
yikes, complicated (files with $ in their name)

PR #394, opened 2023-04-22
feat(upload_files): upload files without file extension

valearna:
In case we want to upload files without file extension - This needs to be discussed first


valearna:
Discussed and approved this new feature

sweng66 (approved):
Looks good! Thanks @valearna

PR #395, opened 2023-04-28
adding code to handle the licenses added by a curator

sweng66:


azurebrd (approved):
Oh, right, users have emails, nice.  Nice, @sweng66

sweng66:
Thanks for reviewing the code and for the sql you suggested during face to face time. @azurebrd

PR #396, opened 2023-05-02
adding code to handle special characters properly and to fix a bug in…

sweng66:
… Comments/corrections curie/pmid assign part


sweng66:
Thanks for your checking/testing !! @azurebrd

PR #398, opened 2023-05-04
fix(tests): run-test-bash and run-functest do not delete all postgres dbs but only literature-test

valearna:


azurebrd (approved):
I don't really understand it, but cool

sweng66 (approved):
@valearna "docker-compose --env-file .env.test run --rm dev_app sh tests/init_test_db.sh"
what does "--rm" mean? delete the container after restarted database?

valearna:
@sweng66 exactly, it means remove the container when it's done. I'm using it only to recreate the test db, so that if you have other dbs in the same postgres docker container they are not deleted

sweng66:
@valearna I see! That is great.

PR #399, opened 2023-05-06
fixed to handle special characters from PubMed and MODs

sweng66:


azurebrd (approved):
Great, thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #403, opened 2023-05-09
Scrum 2869 - fix bugs with file upload and deletion

valearna:
- removing temp files when final file is uploaded
- file deletion now removes mod links first and removes file from db and s3 only when there are no more mods associated


azurebrd (commented):
Thanks @valearna and @sweng66 .  The code makes sense, but I'm sure I'm not following all the subtleties you're talking about on Slack, so I'll defer to Shuai.

valearna:
I tested all cases with the debugger. This is ready for a final review @sweng66

sweng66 (approved):
Looks great! Thanks for taking care of this tricky problem !! @valearna

PR #404, opened 2023-05-09
Scrum 2760 fix special char

FB-ping:


sweng66 (commented):
@FB-ping overall the code looks good! Thanks. But I have two things to point out.

1. mesh_detail_results = db_session.execute(
            " select  qualifier_term, heading_term,  mesh_detail_id from mesh_detail m  where   qualifier_term is not null")

** Looks like you only retrieve mesh data for those rows with qualifier_term so you will miss all the mesh terms without a qualifier. Am I missing anything?

2. if heading_term:
                heading_term_unescaped = html.unescape(heading_term)

** You don't need to check if there is heading_term since it should be always there.

sweng66:
" select qualifier_term, heading_term, mesh_detail_id from mesh_detail m where qualifier_term is not null"

** also you don't need "m" here.

sweng66:
Also I did not see any code to fix the data in Resource table. Is there no special characters in Resource table?

sweng66:
Thanks @FB-ping for fixing the code.

" select  qualifier_term, heading_term, mesh_detail_id from mesh_detail where qualifier_term is not null or heading_term is not null")

You can change this SQL query to 

"select  qualifier_term, heading_term, mesh_detail_id from mesh_detail"

This will return all rows in the mesh_detail table. Every row should have heading_term, some rows also have qualifier_term

unless I missed something

sweng66:
We can chat on zoom if needed @FB-ping

sweng66 (approved):
Looks good! Thanks for fixing the code @FB-ping

azurebrd:
Resources come from an NLM file and the mods, instead of pubmed XML, so I guess those sources just don't have troublesome characters now.

sweng66:
Good to know! @azurebrd

PR #405, opened 2023-05-11
adding formatting sections back to abstracts

sweng66:


azurebrd (commented):
@sweng66 Sometimes there was html markup in the xml, so I couldn't use XML parsers and had to use custom regex, but if you think it works well, then cool.  I don't recall any examples that didn't work, I guess it's already extracting from within <Abstract>(.+?)</Abstract> so it should be fine ?

sweng66:
Thanks for checking the code out!! @azurebrd
 
yes, using XML parser is easier to extract attributes, but extracting the text directly will lose html tags, even losing everything after first tag

so I have to use both XML parser and custom regex to extract the "Label" attribute name and to preserve the html tags.

See the code:

https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-2844/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/xml_to_json.py#L548
 
Also see the abstracts in the following two examples papers from literature_4006, there is one or more <i>*</i> in abstracts:

https://dev4006-literature-rest.alliancegenome.org/reference/AGRKB:101000000389559

https://dev4006-literature-rest.alliancegenome.org/reference/AGRKB:101000000513561

https://dev3006.alliancegenome.org/Biblio/?action=display&referenceCurie=AGRKB:101000000389559

https://dev3006.alliancegenome.org/Biblio/?action=display&referenceCurie=AGRKB:101000000513561

azurebrd (approved):
Cool cool, if you think this is good, then that's great !  Thanks @sweng66

sweng66:
Thanks @azurebrd

PR #407, opened 2023-05-12
Scrum 2857 - fixing minor bugs in referencefile upload api

valearna:
code was breaking when PMC data was already in the system. Excluding PMC from check doesn't change the logic and simplifies things


sweng66 (approved):
Looks good! Thanks @valearna

valearna:
thanks @sweng66

PR #409, opened 2023-05-15
fixed the author/reference triggers so they will be firsed only certa…

sweng66:
…in columns got changed

### DIFF
--- a/agr_literature_service/api/triggers/reference_sql_func_triggers.py
+++ b/agr_literature_service/api/triggers/reference_sql_func_triggers.py
@@ -11,15 +11,42 @@ BEGIN
   END;
 $$;
 """
-reference_update_trigger = """
-DROP TRIGGER IF EXISTS reference_citation_trigger on public.reference;
-CREATE TRIGGER reference_citation_trigger
-AFTER UPDATE OR INSERT ON reference
-        FOR EACH ROW
-        EXECUTE FUNCTION public.reference_update_citation();
+
+# reference_update_trigger = """
+# DROP TRIGGER IF EXISTS reference_citation_trigger on public.reference;
+# CREATE TRIGGER reference_citation_trigger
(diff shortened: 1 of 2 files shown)


valearna (approved):
This is a reasonable solution, at least for now

agr_literature_service/api/triggers/reference_sql_func_triggers.py
```diff
@@ -11,15 +11,42 @@
   END;
 $$;
 """
-reference_update_trigger = """
-DROP TRIGGER IF EXISTS reference_citation_trigger on public.reference;
-CREATE TRIGGER reference_citation_trigger
-AFTER UPDATE OR INSERT ON reference
-        FOR EACH ROW
-        EXECUTE FUNCTION public.reference_update_citation();
```
sweng66, line 19: Thanks both for looking over the trigger code! @valearna @azurebrd

Good question! @azurebrd 
Since current version only handle these two cases (see attached) so I was assuming so.

azurebrd: cool, thanks @sweng66 !

sweng66:
Thanks for looking over the code. @valearna @azurebrd

Good question about deleting a reference @azurebrd! Since original version (current version) only handles UPDATE/INSERT. Seems like we also need to add some thing to the function to handle this situation. Delete the citation if a paper got removed?

Anyway, will leave this to next version! Thanks and good catch! @azurebrd

sweng66:
We can add a reference delete trigger now, but we also need to add code to the function to handle this special case. Well, we don't want to touch the function in this sprint so will leave this reference delete case to later.

azurebrd:
Yeah, if it wasn't there before, it sounds good to leave it.  Thanks @sweng66 !

PR #410, opened 2023-05-17
Scrum 2665

valearna:
- new DB model
- modified API
- modified tests


valearna:
Needs to generate alembic revision. Waiting for #411 to get merged to get the latest alembic file

azurebrd (approved):
Thanks @valearna !  I imagine this is mostly what you showed us yesterday, plus the alembic file, so I haven't looked through it again too carefully.

valearna:
Thanks @azurebrd. Yes, I've added a couple of crud functions that were missing yesterday, but the overall structure hasn't changed

sweng66:
https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-2665/agr_literature_service/api/crud/topic_entity_tag_crud.py#L36

This is not a big deal, but if we use "filter_by", we can save some typing, I meant we can get rid of all these "TopicEntityTagModel"

sweng66:
ignore my comment since majority of codebase use "filter" instead of "filter_by"..

sweng66:
https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-2665/agr_literature_service/api/crud/topic_entity_tag_utils.py#L26

db.query(TopicEntityTagSourceModel).filter(
        TopicEntityTagSourceModel.topic_entity_tag_id == topic_entity_tag_source_id).one_or_none()

Assume TopicEntityTagSourceModel.topic_entity_tag_id is the primary key for "topic_entity_tag",  is the "topic_entity_tag_source_id" the primary key for the source table (topic_entity_tag_source)?

sweng66:
@valearna Overall the code looks great! Thanks.

sweng66 (dismissed):

valearna:
I always prefer `filter` instead of `filter_by` because mypy and flake8 can check the column names

valearna:
They do with `filter`, but in `filter_by` you can put any name and it's not directly linked to the model so they won't know

valearna:
It's totally fine to use `filter_by`, it's only a matter of personal preference

sweng66 (approved):
Looks great! Thanks @valearna

PR #411, opened 2023-05-17
adding new column 'first_initial' to author table and backfilling the data in this new column

sweng66:
* adding new column to API
* fixing several lit processing util functions to handle this new column (taking care of "in" and "out" data)
* adding new column to unit tests
* adding oneoff script to back loading the data (from PubMed) into the new column
* adding a script to load the 'first_initial' for authors if the 'first_initial' is not available in the database but the data is available in PubMed (can run it periodically)


agr_literature_service/lit_processing/utils/db_read_utils.py
```diff
@@ -200,7 +211,7 @@ def get_mesh_term_data(db_session, mod, reference_id_list, query_cutoff):
         for index in range(50):
             offset = index * mesh_limit
             rs = db_session.execute(
-                "select md.reference_id, md.heading_term, md.qualifier_term from mesh_detail md, mod_corpus_association mca, mod m where md.reference_id = mca.reference_id and mca.mod_id = m.mod_id and m.abbreviation = '" + mod + "' order by md.mesh_detail_id limit " + str(
+                "select md.referenfce_id, md.heading_term, md.qualifier_term from mesh_detail md, mod_corpus_association mca, mod m where md.reference_id = mca.reference_id and mca.mod_id = m.mod_id and m.abbreviation = '" + mod + "' order by md.mesh_detail_id limit " + str(
```
valearna, line 214: This looks like a typo

valearna (changes requested):
The PR looks good, thanks @sweng66! I haven't checked the oneoff script in detail but all the changes to the other files look great. There's a typo that need to be fixed

sweng66:
@valearna Thanks for checking out the code! Just fixed the typo (good catch!).

azurebrd (approved):
Looks good to me, @sweng66 .  The key for mapping to first initial could be just the reference_id + order, but you're using the name too just as an added safety ?  Does the PubMed update script need to run first in case PubMed updated the authors between the last run, and this script downloading new xml to get the initials ?  This would be easier than getting the xml from s3, right ?  I think the PR as-is is fine, like, the amount of authors that would be different since the last PubMed update is probably zero or small.  Downloading from s3 would probably be more hassle.

sweng66:
@azurebrd Yes, in theory, we can just use reference_id plus author order as a key, but just want to make sure we have everything match up (both name and order match up between database and PubMed so we get first_initial right). 

Oh, good question. I never thought of getting xml from s3 before. Always getting fresh xml from pubmed. It should be faster downloading the xml from s3. 

I did run pubmed update against database first (for updating the abstracts) and loaded the first_initial right after. Hopefully, everything should match up.

Thanks for looking over the code!!

sweng66:
@azurebrd Do you know how to resolve the "1 change requested" issue? I have fixed the typo.

azurebrd:
Hrm, no, usually when changing the code fixes it, there's an option to "Mark as Resolved"  I don't know why we don't see it in the PR

Getting the XML from PubMed makes sense, we're so used to doing it that way, it has less chance of messing something else up :)

valearna (approved):

PR #412, opened 2023-05-18
fix(topic_entity_tag): fixed errors - use one_or_none instead of scal…

valearna:
…ar (damn chatGPT)


azurebrd (approved):
Thanks @valearna !

PR #414, opened 2023-05-19
SCRUM-2640 Date Ordering

pjhale:
Order by field passed into API.  Still only returns 25 results.  Moved limit to outside subquery... affects performance but is necessary for the ordering clause.

Needs UI PR SCRUM-2640


valearna (approved):
thanks @pjhale

PR #416, opened 2023-05-19
Scrum s673 add citation search dropdown

FB-ping:


azurebrd (commented):
Makes sense, but I don't really know how search works, so I'll defer to Paul and Valerio.

valearna (approved):
Looks good. Thanks @FB-ping

PR #418, opened 2023-05-22
fixed to handle multiple same type of comments/corrections for a paper

sweng66:
eg, two erratumIn papers for a main paper


azurebrd (approved):
Great, thanks @sweng66 !

PR #422, opened 2023-05-26
Adding Filter Param

pjhale:
Splits out subquery into two different ones based on if unfiltered (the  default) or filtered results are wanted.


pjhale:
Currently DRAFT... do not merge.

PR #423, opened 2023-05-26
Scrum 2669 - fixed API to map curies to names using A-team api

valearna:
- Mapping API endpoint now works with new model
- API now returns curie->name mappings for entities and ATP nodes (topic and entity_type)
- Added test (webtest only, not part of PR testing)


azurebrd (approved):
Thanks @valearna !

PR #424, opened 2023-06-01
adding a new endpoint for generating the tracker table

sweng66:
also did the following:
* moved the reusable SQL code out from 'missing_files' function  
* fixed the 'missing_files' function to call this new sql function


pjhale (approved):
Already tested... looks good.

PR #426, opened 2023-06-02
Scrum 2912 sort faster

FB-ping:


valearna (approved):
@azurebrd @sweng66 and I talked about this and we think that the code can be merged as is. We'll create a new task to optimize the code and we'll organize a tutorial on how to optimize queries for performances when I'm back from vacation

PR #429, opened 2023-06-02
add oneoff script to backfill main PDFs from PMC for papers with a PMC package

sweng66:
and fixed to pass 'all_access' for main PDFs from PMC to 'cleanup_temp_file'


azurebrd (approved):
Cool, didn't know about SELECT 1

PR #432, opened 2023-06-09
fixed DQM reference loading script to set MOD curie status properly

sweng66:
and to move mod papers in and out corpus.

Now all papers with a valid MOD curie (is_obsolete = False) should be inside the MOD corpus. The papers with an invalid MOD curie (is_obsolete = True) should be outside of corpus. All papers in current MOD reference files (dqm files) should have a valid MOD curie and should be inside the MOD corpus except those papers with a DOI, but no PMID in MOD files (this set will be taken care of in #2985)

### DIFF
--- a/agr_literature_service/lit_processing/data_ingest/dqm_ingest/sort_dqm_json_reference_updates.py
+++ b/agr_literature_service/lit_processing/data_ingest/dqm_ingest/sort_dqm_json_reference_updates.py
@@ -32,7 +32,8 @@ from agr_literature_service.lit_processing.utils.db_read_utils import \
     get_references_by_curies, get_curie_to_title_mapping, get_mod_abbreviations
 from agr_literature_service.lit_processing.data_ingest.utils.db_write_utils import \
     add_cross_references, update_authors, update_mod_corpus_associations, \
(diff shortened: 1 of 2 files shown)


azurebrd (approved):
Great, thanks @sweng66 !  I didn't know you could subtract sets, that's great.
So there's a section for move_mod_papers_into_corpus to move papers that were false/none into corpus, and move_obsolete_papers_out_of_corpus to move obsolete xref to be outside corpus.  For xrefs that weren't in there before, they're handled by other parts of the script that existed before ?  And mod_curies_not_in_db = mod_curie_set - mod_curie_set_in_db   mod_curies_not_in_db just gets logged but doesn't need to be processed ?

sweng66:
@azurebrd Thanks for checking out code so quickly!!

> For xrefs that weren't in there before, they're handled by other parts of the script that existed before ?

Yes, in theory, everything should be handled by the script. So in the end, just want to fix the missing part by the main script.

> And mod_curies_not_in_db = mod_curie_set - mod_curie_set_in_db mod_curies_not_in_db just gets logged but doesn't need to be processed ?

Yes, for now, just for logging purpose to see if we missed loading some data. These missing ones will be taken care of in #2985

azurebrd:
Great, thanks @sweng66 !

PR #434, opened 2023-06-12
Scrum 2792

azurebrd:


sweng66 (dismissed):
@azurebrd Looks great! Thanks.

One minor thing: can you move commit sections (line 75-78, line 88-91) to after line 93? This way, we can save a few lines of coding =).  Also you can use batch_counter % batch_size == 0 to check if it is reaching batch_size again.  Again, this is one-off script so it is good to merge into main. Thanks Juancarlos!!

azurebrd:
Oh yeah, that's great, @sweng66 !  I've implemented the improvements, thanks !

azurebrd:
Oh yeah, right now there's no updating, but when it gets run in the future, the data might have change, and it might do updates.  I think the curators won't want to run this until WB is ready to switch over Biblio SoT, so it will be a while.

sweng66 (approved):
@azurebrd Thanks for fixing it so quickly. 

Feel free to ignore, but you can remove line 86 now.

azurebrd:
@sweng66 Ah, thanks, good point !  I tried to remove it now, but then flake8 complains about indented if block that doesn't do anything, and if I try to comment out the if statement, it complains about batch_size not being used, so I'll just leave it as-is for now.  Thank you !

sweng66:
I see! Not a big deal. A good place holder!

PR #435, opened 2023-06-15
Scrum 2973 - sort topic and entity tags with null values

valearna:
- added special case to sort null values in entity and topic tags
- for now, sorting still uses a hardcoded map to convert curie to names. I'm waiting for the A-team to enable admin tokens (this will be discussed at a future dev coordination meeting) to convert this map into a dynamic one fetched from the A-team api
- fixed a bug: now the api does not break when the A-team api returns empty result sets

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -140,10 +140,13 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1, p
         return query.count()
     else:
         if sort_by:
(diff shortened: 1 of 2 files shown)


azurebrd (approved):
That's great, thanks @valearna .  It looks more complicated than I would have expected (meaning it's good, and I wouldn't have known how to do it)

valearna:
Thanks Juancarlos. I tried an easier solution (assign 'aaa' or 'ZZZ' to None and use the same sorting mechanism for the in place for the other values) and it didn't work, so I had to find a more complicated solution

PR #439, opened 2023-06-21
Scrum 2766 - create temporary elasticsearch index while rebuilding debezium

valearna:
create a temp index and initially connect debezium to it. After ~5h, remove the connector to the temp index, delete the main index and recreate it, copy the temp index into the main one and connect debezium to the main index. The temp index is not deleted and can be used as a backup until the next restart.


valearna:
I'll deploy this after the review

azurebrd (approved):
That's great, thanks Valerio.  Just curious, what's the PUT for ?

valearna:
The PUT is to "reindex" documents, meaning copy them from the temp index to the main one

azurebrd:
Ah, cool, thanks

PR #441, opened 2023-06-22
Scrum 3018 - new constraints on xref table to allow inserting the same curie multiple times on specific conditions

valearna:


azurebrd (approved):
Thanks @valearna and @sweng66 !

PR #442, opened 2023-06-26
feat(cross_reference): modified alembic revision to not remove the index on xref curie

valearna:


sweng66 (approved):
Looks great! Thanks @valearna

PR #443, opened 2023-06-26
fixed to commit every 50 papers at a time instead of commiting everyt…

sweng66:
…hing in the end


azurebrd (approved):
Great, thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #444, opened 2023-06-27
Scrum 3018 - access xrefs by curie (for exact query) or id (for updates)

valearna:
The xref curie field is not unique anymore as two or more references can be associated to the same curie provided that the curie is not obsolete only for one ref. Therefore we need to use the cross_reference_id field to access xrefs unambiguously. For the "query by exact id" function in the biblio edit UI we allow retrieving an xref by curie but we retrieve the is_obsolete=False first, if any. If all the curies are obsolete, we return the first one.


sweng66 (approved):
Looks great! Thanks @valearna

azurebrd (approved):
Thanks @valearna !

PR #450, opened 2023-07-01
fix(debezium_setup): remove replication slot for citation table when restarting debezium

valearna:
- ES was not building because debezium couldn't create a new replication slot for the citation table. This PR deletes the replication slot from postgres before starting debezium


sweng66 (approved):
Looks great! Thanks @valearna

valearna:
Thanks @sweng66 !

PR #452, opened 2023-07-06
refactoring report utils module

sweng66:


azurebrd (approved):
@valearna I thought we liked standardizing things, why should each logger be its own thing instead of a shared general logger ?

valearna:
@azurebrd if you get a logger for each module, the logger reports the module name so that you know where the log message is generated. This is a standard practice in Python and the logging system remains centralized. Passing the logger to other modules makes things more complicated to understand and manage

azurebrd:
@valearna Ah, gotcha, I thought each script's logger would look at its own self, didn't realize passing the variable around would then refer to the original script.  Thanks.

PR #453, opened 2023-07-06
fixed to set logger in report_utils module so no need to pass the log…

sweng66:
…ger from other programs


valearna (commented):
It seems that some modules have `logger = logging.getLogger()` without name. Can you make sure all modules define their logger with `logger = logging.getLogger(__name__)` @sweng66 ?

sweng66:
Sure @valearna

sweng66:
@valearna fixed =)

valearna (approved):
Cool, thanks @sweng66. It looks great!

sweng66:
Thanks for looking over the code! @valearna

PR #456, opened 2023-07-06
Third try dealing with pycharm commits.

azurebrd:


sweng66 (commented):
why are we only adding entity_type where source = 'alliance'? @azurebrd

azurebrd:
@sweng66 That's what the story is for.  We're currently looking up the curie to name mapping from A-team for any entity, but A-team only has entities when the entity_source is 'alliance'.  When we add the WormBase entities, they'll be curies, but if we try to look them up against A-team, we'll get a 500 error and the API will break.

sweng66:
Oh I see =) @azurebrd

sweng66 (approved):

PR #461, opened 2023-07-07
fixed to handle both DOI and PMCID in PubMed search report

sweng66:


azurebrd (approved):
Great, thanks @sweng66

sweng66:
Thanks @azurebrd

PR #465, opened 2023-07-11
Scrum 2895

ianlongden:
Creating this as a draft for now until we are sure of the exact format for the short_citation.


azurebrd (commented):
I don't really understand triggers, so not approving so someone else also looks at it.  Everything else makes sense though.  I guess -- is to comment out, and it's left in the code for convenience in the future ? (I like to do that).  Or does it mean something functional ?

ianlongden:
-- in the sql scripts comments out code. Yep, left in incase I need to debug or people want to uncomment and see what is going on.

sweng66 (approved):
Looks good Thanks @ianlongden

PR #470, opened 2023-07-12
-r allows for random refs, -n allows naming of the new database, Also…

ianlongden:
… _seqs tables are now queried and updated rather than hard coded


azurebrd (approved):
Cool query for the sequences !
@ianlongden Could the randomize randomly get the same int twice and add the same reference twice ?  That is, can it hit it twice at all.  And if it does, does it create it twice, or does it only create it once and then the subset is one smaller ?

ianlongden:
It can generate the same id but unlikely and will not cause problems except the sanity check at the end may give an error by 1.

PR #473, opened 2023-07-14
SCRUM-3067 change of subject title. SCRUM-2895 remove checking of fir…

ianlongden:
…st auther, use order always


sweng66 (approved):
looks good! Thanks @ianlongden

PR #475, opened 2023-07-17
remove mesh_detail_version

ianlongden:
Will need stop the api,  alembic upgrade head,  then restart of the api.


sweng66 (approved):
Looks good!Thanks. @ianlongden

PR #476, opened 2023-07-17
Scrum 2860

ianlongden:


ianlongden:
So I only have the code to get the sql and results for this and not a complete ticket but as there are fixing to the docker files it might be worth merging into main anyway?

docker/api.dockerfile
```diff
@@ -9,6 +9,10 @@ WORKDIR /usr/local/bin/src/literature
 
 ADD . .
 
-RUN pip install -r requirements.txt
+RUN pip3 install --upgrade pip
```
valearna, line 12: You mentioned why you need to do this at standup but I didn't understand what the reason was

valearna: Oh, I see the comments on scrum-blueteam. Everything makes sense

valearna (approved):

ianlongden:
Yes, I wrote on several things so not surprised it was missed☺️

PR #478, opened 2023-07-20
fixed to not load papers in the MOD false positive list even though they are in the current dqm files

sweng66:
also to make sure all pubmed papers in the current dqm files got added into database even though there is any XREF ID conflict (they should be there already, just to make sure).

another thing I added in the PR is to get rid of '//' in the DOIs  from ABC before comparing them with the data from MODs in order to make the slack report a bit cleaner for WB.


azurebrd (approved):
Great, thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #479, opened 2023-07-21
Scrum 2860 2

ianlongden:


ianlongden:
NOTE: citations are done via an sql trigger but there is still code in there to do this in the api due to it was needed for alembic upgrades. I have set those methods to not be covered in the coverage and put in error messages if they are ever run. That way we can check logs at some later point and make sure this is correct and then delete these methods as it gives the impression that is where they are generated which is incorrect. Very unlike we would do alembic upgrades to old databases from that far back.

PR #481, opened 2023-07-29
Crontab schedule change

sweng66:


azurebrd (changes requested):
Thanks @sweng66 .  Did you say it should happen every day ?  I see the time change, but I thought the 0 8 * * 6 python3 should be 0 8 * * * python3 to happen every day.

sweng66:
@azurebrd Thanks for checking this out.

No, the pubmed update only happens once a week (on saturday) and we don't want to change that. 

While the DQM loading script is running every night so on Saturday these two scripts were running simultaneously and overlapped for a 45-minute interval. That is the reason I moved the pubmed update script from 9am to 8am.

azurebrd (commented):
Ah, thanks @sweng66 !  When I saw the "Files Changed" tab, it only had the change to the pubmed script, but I saw the comment for the commit about the dqm change.  Now I see that change in https://github.com/alliance-genome/agr_literature_service/pull/481/commits/4abd6298d3d0254059c310ff3f585e418946c2f3  but I don't know why it isn't part of the "Files Changed".  Thanks !

azurebrd (approved):
Meant to approve, oops

sweng66:
Oh, I see. @azurebrd

PR #483, opened 2023-08-03
Scrum 3095 - new api to get list of reference with referencefile ids to be processed by textpresso

valearna:


sweng66 (dismissed):
Looks great! Thanks @valearna 

BTW, do we need "source_is_pmc" in line 772 and 776 in reference_crud.py?

azurebrd (dismissed):
Cool, didn't know you could just r.curie > something, I guess it's because it's 0 padded.  Thanks @valearna !

sweng66:
@azurebrd we can also order by title. basically any strings

azurebrd:
*nod*  thanks @sweng66 :)

valearna:
Thanks @sweng66 and @azurebrd. You are right @sweng66 we don't need that data structure anymore. I removed it

valearna:
@azurebrd the curies follow alphanumeric order, so we can use string comparison to know if a curie is larger than another. All new curies will come after the old ones and search is O(logn) since the field is indexed with a btree.

azurebrd (approved):
Thanks both

sweng66 (approved):
Thanks @valearna

PR #485, opened 2023-08-09
refactored the PubMed update programs to use PubMed daily update files as data source 

sweng66:
also added a few checks for adding/updating XREFs
and a function to sort PMIDs into mod buckets.


azurebrd (approved):
Thanks @sweng66 !

azurebrd (approved):
Great, thanks @sweng66 !

PR #486, opened 2023-08-09
Scrum 2764 dump production

FB-ping:


azurebrd (commented):
@FB-ping I'm not sure we need the delete_file_from_s3 function, since it's just used to delete the file from 7 days before, and if using @sweng66 's lifecycle setting on s3, it should happen automatically, but maybe it's good that the function exists so it can be used for future things ?  Same with file_exist_from_s3 being part of the same functionality.  I think it's fine, but it might be confusing to have 2 ways of doing things (one through python, another through s3 lifecycle).

The python stuff makes sense to me, but I don't know about the crontab or Makefile, so I'll defer to Shuai and @valearna (who'll be back soon, and I'd welcome his feedback on the previous paragraph).

Thanks Pinglei !

sweng66:
I agree with Juancarlos. We should use AWS lifecycle to clean up the old files.

FB-ping:
@azurebrd @sweng66 I have no idea that we have the s3 7 day's lifecycle thing, if that, yes, we can use the s3 lifecycle mechanism.   I will just keep the those two functions (delete_file_from_s3 and file_exist_from_s3) but no need to call here.  Just curious how we control the 7 days lifecycle, since we will have files in the monthly much older than 7 days.

sweng66:
@FB-ping setting 7 days lifecycle for the daily dumps only.

azurebrd:
@FB-ping I think it works on a per-bucket basis, which is part of why we made separate buckets for the 3 types of dumps.

FB-ping:
@sweng66 @azurebrd remove the delete/check 7days file now.

azurebrd (commented):
Thanks @FB-ping !  I like having functions that would have a use in the future, but I think @valearna doesn't, so I'd like to see what he says.  Also, there's a warning about the postgres password going into a log, and I don't know how bad that is.  Maybe it would end up in the logs somewhere in github, and that would be bad ?

valearna:
We should already have functions to work with s3 (upload, download, delete) in the repo under api/s3 or lit_processing/utils/s3_utils.py, so I would drop the new functions unless they provide something that is not already there

azurebrd:
Oh, that makes sense, thanks @valearna .  I see the files at api/s3/upload.py and api/s3/download.py  @FB-ping I agree with Valerio that we don't need that twice, and I also find it hard to find functionality when it's in different files, but it's all so complex that I think the current file organization is better than a giant file.  (or a few giant files)
Valerio, if you could check the Makefile and crontab, that'd be great.  Everything else made sense to me.  And do you think it'd be better to remove the log command that has the password ? (I would think we do)

valearna:
The crontab looks ok but the Makefile is calling python3 directly and this would be the local python installed on the machine where the make command is called. It would be better to use a docker image to run python, possibly the dev_app. We have other commands in the Makefile that are using the dev_app to run scripts. You can take a look at `alembic-create-migration` @FB-ping.

valearna:
Yes, removing the log commands with the password sounds good

FB-ping:
@azurebrd @valearna if no problem with the Makefile change, could you approve the PR so I can merge into main ?

valearna (approved):
Thanks @FB-ping!

PR #487, opened 2023-08-11
Scrum 3045 - new topic entity tag db schema and functions

valearna:
Updated models to reflect the latest class diagram:
 - main tag table now contains all columns related to the tag itself (no shared core tag data anymore)
 - redefined source to map a source object (what generated the tag - e.g., a script, tool, a curator through a specific interface, an author through acknowledge)
 - validation is not pre-computed anymore. Each tag now has a list of tags that validates it (validated_by field in the main tag model)
 - will need a separate story to return validation values for each tag


sweng66 (approved):
Looks great! Thanks @valearna 

one minus thing and a couple of questions for this PR:

* We can get rid of line 38 in topic_entity_tag_crud.py. It is duplicate of line 37.

* Not a big deal. but can we get rid of "db.commit()" in topic_entity_tag_crud.py#58 since there is db.commit() in "validate_tags_on_insertion" function already?

* How does  the "validated_by" column work? Looks complicated!
  Does it store all related entry in this column?

* In patch_data_dict = patch_data.dict(exclude_unset=True)

With "exclude_unset=True" so we only update a column if its value is passed in, right?

valearna:
Thanks @sweng66! I removed the duplicate lines following your comments.
validated_by is a list of tags that validate the current tag and is based on the sqlalchemy design pattern for many-to-many relationships, plus some configuration for self-relationship (it's a list of tags related to a tag) and more config to make it work with sqlalchemy continuum (the version tables). This is the documentation I followed: https://stackoverflow.com/questions/1889251/sqlalchemy-many-to-many-relationship-on-a-single-table https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#many-to-many. Yes, `exclude_unset=True` removes None values and this is useful when patching an object. We use this in other patch methods. We should revisit all the error messages, but I think we have a separate story for that and I didn't want to make this one too complicated.

sweng66:
Cool. Thanks @valearna

sweng66 (dismissed):

valearna:
Changing the db model and API following comments from today's metadata meeting and the new structure in the class diagram

sweng66 (approved):
Looks good! Thanks @valearna

PR #490, opened 2023-08-14
feat(topic_entity_tag): new api to get topic entity tag source by name

valearna:
This is needed by the UI to retrieve the "curator" source if it already exists or create a new source otherwise


sweng66 (approved):
Looks great! Thanks @valearna

PR #491, opened 2023-08-14
fix(topic_entity_tag): validating only for the same mod + added exception handling

valearna:
…tion handling for tags


valearna:
This is a hotfix made after testing the API and UI with @cerivs

azurebrd (approved):
Thanks @valearna

sweng66 (commented):
Looks good! Thanks @valearna

PR #492, opened 2023-08-15
SCRUM-2144 Adding Mod Reference Type

pjhale:
Adding the mod reference type to ES and the API.  Required changes to ksql, crud, and debezium.  Currently, stores all mod reference types in a single bucket without regard to mod affiliation since this was considerably simpler.  Needs the UI PR to display properly.


valearna (approved):
Thanks @pjhale !

valearna (approved):

PR #493, opened 2023-08-15
fix(debezium): fixed typo

valearna:


pjhale (approved):
I must have fixed this in the terminal... thanks for finding that.

PR #494, opened 2023-08-17
feat(topic_entity_tag): add CASCADE on foreign key deletion in topic_entity_tag_validation model

valearna:
Foreign key relationships in the 'topic_entity_tag_validation' table are now set to 'CASCADE' on deletion. This change ensures that when a 'topic_entity_tag' is deleted, all related entries in 'topic_entity_tag_validation' are also removed, maintaining the integrity and consistency of the data. Accompanying Alembic migration to apply the change to the database was created.


sweng66 (approved):
Looks good! Thanks @valearna

PR #497, opened 2023-08-18
feat(topic_entity_tag): Add validation values to topic entity tag output (SCRUM-3187)

valearna:
Added logic to calculate validation values based on data from the validated_by association and include these values in the output of the 'show' and 'list' endpoints. This allows users to easily see validation status from different sources (author, curator, curation tools) for each tag. Extended schema and updated tests to reflect these new fields.


sweng66 (approved):

sweng66:
I added a comment, not sure where it goes

PR #500, opened 2023-08-28
Scrum 3187 - fixed validation by including current tag when calculating validation values

valearna:
If validation_type for the current tag is equal to the provided validation_type, the 'negated' value of the current tag is added to the validation. This covers the case where only one additional tag for the validation_type exists and it has a conflicting 'negated' value with the current tag. The resulting validation value is now None instead of False. This makes validation on conflicting 'negated' values for the same type always None, which is more consistent and easier to interpret.


azurebrd (approved):
Thanks @valearna !

PR #503, opened 2023-08-30
fix(topic_entity_tag): sort by now restricted to elements related to …

valearna:
…the specified reference

Also added mapping from a-team api on any possible value coming from atp. For now, this is restricted to entity_type and topic

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -179,7 +179,7 @@ def show_source(db: Session, topic_entity_tag_source_id: int):
     return source_data
 
 
-def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1, page_size: int = None,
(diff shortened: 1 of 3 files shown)


azurebrd (approved):
Thanks @valearna !  Weird, how did this work before without getting the token ...

valearna:
It was using a hardcoded mapping for genes and alleles only. Now it works on all values in atp for entity_type and topic. We'll need to add different mappings for other columns, or ask the a-team to create a single endpoint to map curies to names. We already asked for this, haven't we?

azurebrd:
Oh, to map the entity_type.  But how did the topic and entity curies work before ?

I don't know whether we've asked them for an any-endpoint to map curies to names.

valearna:
The code I modified is used only for sorting and it was using hardcoded values to convert curies to names for the column to sort. Now I connected it to the function to map curies to name, which was already used to map topic, entity_type and entity curies in the UI from a different api endpoint

azurebrd:
Got it, thanks.

PR #505, opened 2023-08-30
fix(topic_entity_tag): Add field constraints to topic_entity_tag_sche…

valearna:
…mas.py

This change introduces the use of `pydantic` constraints, specifically `Field` and `constr`, to the schema definitions within `topic_entity_tag_schemas.py`. Previously, fields of the schema classes could accept empty strings or exclusively white spaces, which led to improper utilization of the database and reduced data integrity. By placing a minimum length requirement on each string-based field, we ensure all data being pushed to the database is non-trivial and supports our application's use cases. The updated unittest reflects this enhancement by asserting that creating entries with empty strings now results in HTTP 422 status.


sweng66 (dismissed):
Looks good! Thanks @valearna

sweng66 (approved):

PR #507, opened 2023-08-31
feat(topic_entity_tag): Add display_tag to mapping function

valearna:
Map curie to name for display_tag


valearna:
Thanks @azurebrd !

PR #509, opened 2023-09-07
Scrum 2993

azurebrd:


azurebrd:
@valearna @sweng66 I think this works, but we shouldn't merge this into main until the UI is updated to work with it.  But I wanted to get your feedback in case you think some things should work differently.

valearna (commented):
It looks good, thanks @azurebrd! I'll approve it when ready to be merged to avoid merging it by mistake. Do we need mod_curie because everything that is in_corpus must have a mod_xref?

sweng66 (approved):
Looks good! Thanks @azurebrd

sweng66:
oops I approved lol

sweng66:
Should we fix the program to check and make sure both mod_curie and mod_mca are not null or both are null? 
just to make sure all in-corpus papers have a valid mod_curie and all papers with a valid mod_curie must be in given mod-corpus.
@azurebrd @valearna

azurebrd:
Thanks @valearna and @sweng66 !  I think it's fine to be approved.  It's probably fine if it gets merged too, even if the UI doesn't work for creating PMID references on stage, it won't be bad.  And even if it gets into prod, I don't think anyone's doing that.

I think the endpoint requires all the values to be there, and that blank isn't allowed, is that right, Valerio ?  Shuai, do you think it could be a problem if someone tries to use the function from some other script but only adds values to exactly one of mca vs mod_xref ?  I think Ceri wanted to be able to create it as the okta-login mod (for the mca), but assign the xref to a different mod's curie.  If that's bad, we should discuss it with her.

sweng66:
@azurebrd I don't think there is any other script using this function. 

Assign a mod_curie to a paper that is in another mod corpus is fine, but the newly added cross_reference row will be updated (set to invalid) by dqm loading script through cron. Also the mca row will be updated as well (move it out of corpus) unless curators have done something before the cronjob run. 

Am I missing anything?

valearna:
@azurebrd I think that the values from the router are nullable so we should check if any of the three is null and we should return an error. Maybe we can add this to the tests as well?

azurebrd:
@sweng66 cool cool, it's just 2 tests and the functional test that uses this function.  @valearna + @sweng66 I guess we should require all fields to have non-null values.  @valearna I'm not sure how to make the API return an error, but I'll try / ask you if it's not clear.  Adding it to the tests sounds good.

@sweng66 I've written to @cerivs about the issue, and we can discuss at / after standup.  Maybe we just require the oktaMod to be the same mod as the xref mod_curie prefix.

sweng66:
@azurebrd Good idea ("require the oktaMod to be the same mod as the xref mod_curie prefix.") !

valearna:
Yes, let's discuss how to manage null values @azurebrd. At this point we should also consider using a json data request object (instead of separate query parameters like /add/{pubmed_id}/{mod_curie}/{mod_mca}/) and validate it against a schema that requires non-null and possibly non-empty strings and returns errors automatically.

PR #510, opened 2023-09-07
Add input validation for MOD curie in the reference router

valearna:
This commit adds checks to verify the input format of MOD curie in the reference_router.py file. It enforces a strict format for MOD curie that it must contain a character ':' exactly once and neither of its prefix nor id can be an empty string. HTTP exceptions have been added to throw errors when these conditions are not met. This validation is necessary to ensure data integrity and prevent potential errors caused by invalid inputs.


azurebrd (approved):
Thanks @valearna !

sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):
Yes, thanks @valearna !

PR #519, opened 2023-09-28
SCRUM-3047-General-Topic-Validation

pjhale:


pjhale:
The test test_show_reference_referencefiles_referencefile_mods randomly failed in Valerio's last commit for no apparent reason... seems to be fixed now.  Working on adding some additional tests.

azurebrd:
There were some tests that used to fail sometimes because some transactions weren't happen in the same order, and those were rewritten, but it's possible some tests are still doing that.  But also some tests just fail sometimes and there's no apparent reason, like you said.

valearna:
I'm trying to fix the error with tests @pjhale

pjhale:
thanks Valerio... moving this to ready for review.

valearna (approved):

PR #521, opened 2023-10-04
adding two parameters to  'show_all_reference_tags' to retrieve all species associated with a given paper

sweng66:
and also returning all reference tags associated with one or more species if they are passed in for a paper.

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -233,7 +233,7 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, token: str, page
     if species_only:
         species_list = db.query(TopicEntityTagModel.species).filter_by(
             reference_id=reference_id).distinct().all()
-        distinct_species_list = [species[0] for species in species_list]
+        distinct_species_list = [species[0] for species in species_list if species[0] is not None]
         return jsonable_encoder(distinct_species_list)
 
     query = db.query(TopicEntityTagModel).options(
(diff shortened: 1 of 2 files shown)


agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -219,16 +219,29 @@ def show_source(db: Session, topic_entity_tag_source_id: int):
...
     if page < 1:
         page = 1
     if sort_by == "null":
         sort_by = None
     reference_id = get_reference_id_from_curie_or_id(db, curie_or_reference_id)
+
+    if species_only:
+        species_list = db.query(TopicEntityTagModel.species).filter_by(
+            reference_id=reference_id).distinct().all()
+        distinct_species_list = [species[0] for species in species_list]
```
valearna, line 236: There are cases where species is null (None). Should we change this to `[species[0] for species in species_list if species[0] is not None]`?

valearna (dismissed):
Thanks @sweng66 your changes look good. We will have to generalize the parameter `species_only` to make it work for all other columns on which we want to apply a filter but for now this is good and it shouldn't be complicated to generalize the function later.

agr_literature_service/api/crud/topic_entity_tag_crud.py
sweng66: @valearna Oh good catch!

I will fix that!

sweng66:
@valearna Yes, good point! I didn't realize that we will add filtering for other columns later. Can we work on this in a pair programming session? I think I need help on this (how to generalize the parameter) eg, filter_by=species?

valearna:
yes, pair programming sounds good, but we can merge your PR first as all changes are good

valearna (approved):

sweng66:
Cool. Thanks @sweng66

PR #524, opened 2023-10-09
Scrum 3047 - added oneoff script to remove all referencefiles for a specific mod + added http exception during validation if a-team api is down

valearna:


sweng66 (approved):
Looks good. Thanks @valearna 

I was wondering about the usage for this oneoff script.

valearna:
Thanks @sweng66 I need to upload the latest WB papers as we need to switch to the new WB system that uses pdfs from ABC. Since some older papers have been modified I prefer to delete all WB referencefiles and do the upload from scratch. This script is the safest way to remove referencefiles and the related s3 files if they are not connected to other MODs

sweng66:
@valearna Oh I see. That makes sense!

PR #528, opened 2023-10-16
Fixing Topic/Entity Editor API to handle sorting for translated curies (eg. entity, species, topic, etc)

sweng66:


azurebrd (approved):
I lost my review on code-change (I'd have thought github would keep my comment somewhere, but I don't see it).  Basically :

Thanks @sweng66 .  Some of this code is more @valearna style, so I don't understand it, but trust that it makes sense to the two of you, and that it does what you want it to.  Also there's a token in there, so it might conflict with what @ianlongden is working on, but not sure.

sweng66:
Thanks @azurebrd !

valearna (approved):
Thanks  @sweng66 !

PR #530, opened 2023-10-18
SCRUM-3081

pjhale:
Adding ability to negate pubmed publication status.


valearna (approved):

pjhale:
> Thanks @pjhale . It'd be good to have unit tests, so that it doesn't drop our coverage, but I don't know if adding this code has dropped our coverage

I don't think it should... in the API we aren't adding any new functions just adding a new parameter to an existing function.

azurebrd:
Gotcha, thanks @pjhale

PR #531, opened 2023-10-18
Scrum 3106

azurebrd:


sweng66 (approved):
Looks good! @azurebrd 

so we can't create a WB curie for a paper that is already in the database with an invalid MOD curie. right?

azurebrd:
> Looks good! @azurebrd
> 
> so we can't create a WB curie for a paper that is already in the database with an invalid MOD curie. right?

Oh, that's a good point @sweng66   We didn't talk about that.  I think it's querying for any reference by that mod, and sorting by is_obsolete, but if there were an obsolete one, it wouldn't allow creating one.  I guess we should ask @cerivs and Kimberly ?

PR #532, opened 2023-10-19
A new script for searching PubMed by DOI 

sweng66:
This script does the following:

1. if a PMID is found for a DOI and the new PMID is not currently in the database, this PMID will be added to the database and will be associated with the paper with the given DOI. In the end, the program will update these papers (with a pmid added) with the data from PubMed.
2. If a PMID is found for a DOI and the new PMID is already in the database (associated with another paper), then a pair of duplicate papers is identified - one with a DOI and another one with a PMID. In the end, the program will send a report to a slack channel if one or more pairs of duplicate papers are identified.


sweng66:
Thanks @valearna

PR #538, opened 2023-10-25
fixed code to ignore any papers with prepublication_pipeline = True when doing data integrity checking/cleaning…

sweng66:
This means any papers with prepublication_pipeline = True and with MCA corpus = True will stay inside corpus even though they don’t have a valid MOD ID (curie).


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #539, opened 2023-10-26
add_novel_topic_data_to_topic_entity_tag

ianlongden:


sweng66 (approved):
Looks good! Thanks @ianlongden 

A minor question: do we need to set unique=False in toptic_entity_tag_model.py?

ianlongden:
> Looks good! Thanks @ianlongden
> 
> A minor question: do we need to set unique=False in toptic_entity_tag_model.py?

Technically no, unique = false is the default, but by stating it it, it is just more obvious.

sweng66:
OK I see. Thanks @ianlongden

azurebrd (approved):

PR #540, opened 2023-10-26
adding check to make sure the okta user have permission to delete the Tet row

sweng66:


azurebrd (approved):
Oh, there was already oktaAccess.ALL_ACCESS, that's great.

Weird about mypy...
Might be good to leave some commented out code that it should be 403, but it's 404 because mypy ?  Wonder how that passed mypy before, if this wasn't changed.

Thanks Shuai.

sweng66:
@azurebrd Sounds great to leave the old code (but comment it out)

azurebrd (approved):
Thanks @sweng66

sweng66:
Thanks! @azurebrd

PR #542, opened 2023-11-02
dev_app needs ATEAM_API_URL to start api from inside dev_app with Ian…

azurebrd:
…'s changes requiring some variable


sweng66 (approved):
Thanks @azurebrd

PR #545, opened 2023-11-03
Scrum 3103

azurebrd:


sweng66 (dismissed):
Looks good! Thanks @azurebrd

tests/api/test_search.py
```diff
@@ -188,12 +188,6 @@ def test_search_on_abstract(self, initialize_elasticsearch, auth_headers): # noq
             res = client.post(url="/search/references/", json=search_data, headers=auth_headers).json()
             assert len(res["hits"]) == 2
 
-    def test_search_need_review(self, test_mca): # noqa
-        with TestClient(app) as client:
-            res = client.get(url="/search/need_review", params={"mod_abbreviation": "0015_AtDB", "count": 10})
-            assert res.status_code == status.HTTP_200_OK
-            assert len(res.json()) > 0
-
     def test_search_sort(self, initialize_elasticsearch, auth_headers): # noqa
```
valearna, line 191: Do we need to move this test to the new file?

valearna (approved):
Thanks @azurebrd, search and sort functions in separate files and routers look much better. There's a sort test function that is still in the search tests and should be moved to the new file, but it should be an easy fix

PR #549, opened 2023-11-10
fix(api): add resource descriptor obj to the db inside the for loop

valearna:
This fixes a bug for which only the last obj in the resource descriptor yaml file was loaded


azurebrd (approved):
d'oh, python ...  thanks @valearna :)

valearna:
:smile: Thanks @azurebrd !

PR #550, opened 2023-11-10
Scrum 3497 - new API endpoint to read xrefs in bulk using list of curies

valearna:
This reduces the number of calls to the xref API that were blocking the UI while loading the results in the search page


azurebrd (approved):
Thanks @valearna !

valearna:
thanks @azurebrd !

PR #553, opened 2023-11-21
Upload 6 pdfs for stage papers

sweng66:


azurebrd (approved):
Thanks @sweng66 !

I would have thought that upload_pdf_file_to_s3 already existed somewhere and didn't need to be written, just imported.  But I don't actually know.  We don't ever need to upload to prod, just stage, and I see the cronjob does that, which made me think that this might be code that exists somewhere else.  

Thanks !

sweng66:
@azurebrd You are right. We can use the available function!!! Thanks.

sweng66:
wait. did you mean that we already have code some where that only uploads the files to s3 for stage papers? @azurebrd

azurebrd (approved):
No, I don't actually know, @sweng66 .  If the existing function is different than what you need for this task, then using a new function makes sense.  I just thought this function looked like something that already existed, since it also uploaded to prod.  The cronjob only running on stage/build makes sense

sweng66:
@azurebrd I see. Thanks

PR #559, opened 2023-12-05
Autocomplete XRefs endpoint for WB

valearna:
Using DB queries instead of Elasticsearch

### DIFF
--- a/agr_literature_service/api/crud/cross_reference_crud.py
+++ b/agr_literature_service/api/crud/cross_reference_crud.py
@@ -178,3 +178,22 @@ def show_changesets(db: Session, cross_reference_id: int):
             }
         )
     return history
+
+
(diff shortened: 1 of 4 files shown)


sweng66 (approved):
Looks great! Thanks @valearna

azurebrd (approved):
Looks good, thanks @valearna .  When I was working on the Sort endpoint, that was under search, and we changed it to a query, so I thought about moving it to a Query endpoint, but you (Valerio) suggested moving it to a Sort endpoint, which is where it's at now.  Should this be in the cross_reference/ section in a similar way ?

valearna:
Thanks @azurebrd. Yes, moving it to the xref router makes sense

azurebrd (approved):
Cool, thanks @valearna !

PR #560, opened 2023-12-05
fixed dataPath to use full path instead of relative path

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #562, opened 2023-12-05
Improve cross-ref search results limit handling and ordering

valearna:
Made an improvement to the handling of limit results in the cross-reference search within `cross_reference_crud.py`. Now, if matching cross-references count is less than 20, an expanded query is executed, and matching curies are extended with the results. Also, results are ordered by `CrossReferenceModel.curie` to provide a structured display of data.


azurebrd (approved):
Thanks @valearna .  I can't tell what kind of object matching_curies is, but does it filter out getting doubles from the second query that were already in the first query ?  If it's a hash/dict, cool, if it's a list, maybe it would find the same matches again in the top part of the returned data.  (It looks like it's a list/array, so that might be a problem, on the WB case I throw it in a hash, and then output, and I might be doing it wrong, as opposed to the way I described it on the call, with both sets ordered separately)

sweng66 (approved):
Looks good! Thanks @azurebrd

PR #563, opened 2023-12-09
Fixed pagination issues with an additional order_by condition on topic_entity_tag_id

valearna:
Overall query and paginated queries with offset and limit were returning non-overlapping results because of randomness in the returned sets of paginated queries. Added order_by condition on topic_entity_tag_id to avoid random ordering among tags with the same topic and ref curie.


sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):
Thanks @valearna

valearna:
Thanks @sweng66 @azurebrd

PR #568, opened 2023-12-19
Scrum 3568 - t&e tag field 'negated' now accepts null values

valearna:
- modified db model
- modified API schemas
- modified validation algorithm
- added test
- generated alembic revision


azurebrd (dismissed):
Neat that you can union like negated: Optional[Union[bool, None]] = False

valearna:
Thanks @azurebrd for the review and @sweng66 for fixing the unit test error

PR #569, opened 2023-12-20
adding a new column (taxon_ids) into mod table

sweng66:


azurebrd (commented):
I'm not sure how this was supposed to work.  Were we also going to track the default taxon somehow ?

sweng66:
We will need to create an end point to return the mod to taxonIDs mapping and also update the UI to get rid of hard-coded taxon ID to species name mapping and instead to get the mapping from API. @azurebrd

azurebrd:
Sounds good @sweng66 , so this is just one part.  But do we still need somewhere in the database a place to set which taxon is the default taxon ?

sweng66:
It is set when we applied the alembic file @azurebrd

valearna (approved):
Looks good, thanks @sweng66!

sweng66:
Thanks @valearna

PR #570, opened 2024-01-04
scripts for end of year update

sweng66:
plus a fix to db_write_utils.py


valearna (approved):
I don't understand the details but they look good

sweng66:
@valearna Thanks!

PR #572, opened 2024-01-04
Scrum 1807 - added docker-compose to jenkins

valearna:
Also modified jenkins configuration in the UI pipeline on http://dev-ec2.alliancegenome.org:49001 to use docker-compose instead of docker. 

Some of the advantages of using docker-compose in jenkins are:
- simplified script
- build the image before stopping, so less downtime


sweng66 (approved):
Looks good! Thanks @valearna

PR #575, opened 2024-01-11
add test that creates three references, makes relations from first to…

azurebrd:
… second and first to third, then merged second and third, and tests that first to third has both types of relations


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #576, opened 2024-01-16
Scrum 3049 - Validation now works for negative tags

valearna:
- Re-organized validation code
- Added validation for negative tags
- When calculating validation values (on the fly from the API), chain of validating tags are now evaluated to make sure there are no conflicting tags upstream (e.g., a tag is validated positive by a tag that is validated negative by another tag that does not validate the first one)
- Code not completely factorized but left duplicated lines to maximize readability


valearna:
Adding a use case for the discussion at standup: a positive tag on "high throughput phenotype" is validated (as validated_wrong) by a negative tag on "phenotype" (more generic), but it's not validated by a positive tag on "phenotype". The positive and negative tags on "phenotype" are conflicting, but if we look at just the tags directly validating the tag on "high throughput phenotype" the tag it's validated_wrong and not validation_conflict. The code changes in this PR add recursive traversal of validation relationships to address this. Since validation is reference and MOD related, this shouldn't slow down too much validation.

PR #580, opened 2024-01-17
Fixed the topic_entity_tag create_tag end point to check for existing tag and return the tag data (curie+name)

sweng66:


agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -4,13 +4,15 @@
 """
 from collections import defaultdict
 from typing import Dict
```
azurebrd, line 6: extra line ?

agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -31,7 +33,7 @@
 ATP_ID_SOURCE_CURATION_TOOLS = "curation_tools"
 
 
-def create_tag(db: Session, topic_entity_tag: TopicEntityTagSchemaPost) -> int:
+def create_tag(db: Session, topic_entity_tag: TopicEntityTagSchemaPost) -> dict:
```
azurebrd, line 36: oh, how did this work as an int before ?

azurebrd: Oh, it used to be an int, nevermind

azurebrd (approved):
Great, thanks @sweng66 !  (feel free ignore that single comment, I wrote because I wrote what I thought was a useful different single comment, but it wasn't useful)

agr_literature_service/api/crud/topic_entity_tag_crud.py
sweng66: It was returning the newly generated topic_entity_tag_id before so it was int @azurebrd

sweng66: I didn't pay attention to that 'int' earlier until the unittest failed =)

azurebrd: Thanks @sweng66 :)

sweng66:
Thanks @azurebrd

PR #582, opened 2024-01-20
Fixed conversion from sqlalchemy db obj to dict + updated tests to use new tet post return values

valearna:
Calling jsonable_encoder on a db obj with loops in the relationships was returning a max recursion error. The exclude parameter of jsonable_encoder only works if the passed object is in dict format. This looks like a bug in the decoder function. Made a first conversion with vars() and then called jsonable_encoder with the problematic relationship (validated_by) in the exclude list.

Updated tet web tests to use the new format returned by tet create function.

Please remember to run tests locally before creating PRs with changes to tet functions, as most tests are webtests skipped by github actions as they include calls to A-team APIs.

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -432,7 +432,9 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1,
         mod_id_to_mod = dict([(x.mod_id, x.abbreviation) for x in db.query(ModModel).all()])
         all_tet = []
         for tet in query.offset((page - 1) * page_size if page_size else None).limit(page_size).all():
-            tet_data = jsonable_encoder(tet)
+            tet_data = jsonable_encoder(vars(tet), exclude={"validated_by"})
+            if "validated_by" in tet_data:
+                del tet_data["validated_by"]
             add_validation_values_to_tag(tet, tet_data)
             tet_data["topic_entity_tag_source"]["mod"] = mod_id_to_mod[tet.topic_entity_tag_source.mod_id]
             all_tet.append(tet_data)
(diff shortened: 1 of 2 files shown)


sweng66:
So vars(tet) converts the tet object into a dictionary.. Cool.

sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):
Thanks @valearna !  We should make sure to tell the other devs that don't review this PR (Pinglei, but also Ian, and just in case Paul).  How are there loops in our TET data ?  (or is it loops in the ATP ontology, the code looks like vars is on abc tet data)

Why are some things now passing the specific ["topic_entity_tag_id"], when before they weren't ?  Did the format change, or was it not the right thing before ?

sweng66:
@azurebrd I think it is due to the create_tag function change right? @valearna  
now the create_tag returns a dict not topic_entity_tag_id anymore

azurebrd:
@sweng66 Oh, I didn't see changes to create_tag in the PR, so I figured it was still working like it did before ?  (but I don't remember how the whole thing works, when I looked at it before, it seemed really complex and didn't really follow it)

sweng66:
In my last PR, I changed create_tag function to return dict since it has to return the tags that are already in the database. @valearna 

I did not read the tag validation code

azurebrd:
@sweng66 ooooooh, I see, I see, thank you !

valearna:
Thanks @azurebrd and @sweng66. Yes, we need to make sure all devs run local tests when changing tet code. The loops are in the db objects. If two tags validate each other (e.g., a negative and a positive tag on the same topic) they create a loop in the 'validated_by' field and since jsonable_encoder recursively follows all relationships of an object it got stuck in the loop. We didn't really need to return the 'validated_by' objects with this api endpoint, so I excluded that field when calling jsonable_encoder, but there was the bug I described above with that function, so I needed to convert the first level object to a dict first and then pass it to jsonable_encoder in order to make the exclude option work.

sweng66:
Sorry I missed those webtests during my fixes @valearna

valearna:
no worries @sweng66. Unfortunately it's not ideal to have webtests for important features.

azurebrd:
Got it, thanks @valearna and @sweng66 .  It's hard having a bunch of devs working on the same code, where someone approves a change without someone else knowing what's changed.

PR #583, opened 2024-01-20
Corrected validation functions

valearna:
I discovered that some of the positive/negative - more generic/specific combinations for validation were wrong. I restructured the validation functions to fix these cases.


azurebrd (approved):
Thanks @valearna !

PR #584, opened 2024-01-22
scrum-3446 add index on foreign key

FB-ping:
1. add index on created_by/updated_by
2. add index on other missed foreign keys.


sweng66 (approved):
Looks good to me! Thanks @FB-ping

azurebrd (approved):
Thanks @FB-ping and @sweng66 .  I don't really know, but I think it looks good.

PR #585, opened 2024-01-24
fixed so adding duplicate species will work too

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #586, opened 2024-01-26
Scrum 3646 - added TET tests as normal unit tests (not webtests)

valearna:
Added mock functions and set return values


sweng66 (approved):
Looks good! Thanks @valearna 

so you updated get_map_ateam_curies_to_names in "topic_entity_tag_utils.py" to return curie if there is no name or symbol field (eg, geneSymbol) available just for the mock test, right?

valearna:
Thanks @sweng66. Yes, the tests were using ATP:0000122 which has been deleted. The change I added always returns something even if the node does not exist instead of failing

sweng66:
cool @valearna

tests/api/test_topic_entity_tag.py
```diff
@@ -172,7 +173,11 @@ def test_get_all_reference_tags(self, auth_headers): # noqa
 
     @pytest.mark.webtest
```
azurebrd, line 174: @valearna should this still be a webtest, given it mocks ?

tests/api/test_topic_entity_tag.py
```diff
@@ -224,9 +229,12 @@ def test_validation(self, test_topic_entity_tag, test_reference, test_mod, auth_
             response = client.get(f"/topic_entity_tag/{test_topic_entity_tag.new_tet_id}")
             assert response.json()["validation_by_author"] == "validated_wrong"
 
-    @pytest.mark.webtest
     def test_validation_wrong(self, test_topic_entity_tag, test_reference, test_mod, auth_headers, db):  # noqa
-        with TestClient(app) as client:
+        with TestClient(app) as client, \
+                patch("agr_literature_service.api.crud.topic_entity_tag_crud.get_ancestors") as mock_get_ancestors, \
+                patch("agr_literature_service.api.crud.topic_entity_tag_crud.get_descendants") as mock_get_descendants:
+            mock_get_ancestors.return_value = []
```
azurebrd, line 236: In this case, the mock happens to prevent the web lookup, but the values aren't needed so mock_get_... are empty, and that's okay ?

azurebrd (commented):
Thanks @valearna !  Good to see how it should work.  Had a couple of questions in the code as comments.

tests/api/test_topic_entity_tag.py
valearna: Good catch @azurebrd ! I'll create a new PR for this. There is also another test with the same issue.

tests/api/test_topic_entity_tag.py
valearna: Yes

azurebrd: Cool, thanks

tests/api/test_topic_entity_tag.py
azurebrd: Cool, thanks !  Missed that one :)

PR #590, opened 2024-01-31
Scrum 3218 - new API endpoint to revalidate all tags

valearna:
- API endpoint starts background process that for each TET removes the validation entries for tags validated by the current tag and then recalculate validation provided by the tag (one-way, not validation to the current tag by existing tags).
- sends notification email when done
- user can choose whether to delete all validation entries at the beginning of the process or delete only those related to the tag currently processed.


agr_literature_service/api/routers/topic_entity_tag_router.py
```diff
@@ -133,3 +136,40 @@ def show_all_reference_tags(curie_or_reference_id: str,
 def get_map_entity_curie_to_name(curie_or_reference_id: str,
                                  db: Session = db_session):
     return topic_entity_tag_crud.get_map_entity_curie_to_name(db, curie_or_reference_id)
+
+
+def revalidate_tags_process_wrapper(already_running, email: str, delete_all_first: bool, db: Session):
```
azurebrd, line 141: what calls this ?  wondering where the email comes from

valearna: That's a wrapper function that represents a 'task' that the background process can run. The email is passed by the user through the API

agr_literature_service/api/routers/topic_entity_tag_router.py
```diff
@@ -133,3 +136,40 @@ def show_all_reference_tags(curie_or_reference_id: str,
...
+        already_running.value = False
+
+
+@router.get('/revalidate_all_tags/',
+            status_code=200)
+def revalidate_all_tags(email: str,
+                        delete_all_tags_first: bool = False,
+                        user: OktaUser = db_user,
+                        db: Session = db_session):
+    if not user.groups or "SuperAdmin" not in user.groups:
```
azurebrd, line 155: SuperAdmin seems good, but I don't know who's meant to run this.  For example, you and I are, but Shuai and Pinglei aren't.

azurebrd (approved):
Thanks @valearna !

agr_literature_service/api/routers/topic_entity_tag_router.py
azurebrd: Cool, thanks

agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -313,12 +314,32 @@ def validate_tags(db: Session, new_tag_obj: TopicEntityTagModel):
...
-                                     related_tag.topic_entity_tag_source.validation_type is not None]
-    validate_new_tag_with_existing_tags(new_tag_obj, related_validating_tags_in_db)
+    if validate_new_tag:
+        related_validating_tags_in_db = [related_tag for related_tag in related_tags_in_db if
+                                         related_tag.topic_entity_tag_source.validation_type is not None]
+        validate_new_tag_with_existing_tags(new_tag_obj, related_validating_tags_in_db)
     db.commit()
 
 
+def revalidate_all_tags(db: Session, email: str, delete_all_first: bool = False):
```
valearna, line 324: I'm not sure if I should add more db.commit() in this function. What's currently implemented deletes all validation entries from the current tag first, then recalculate the new validation and adds the related entries, then commit (as part of the validate_tags function). If delete_all_first is selected then the first commit will delete all validation entries in the table and then add the first re-calculated entries. Does this sound good @azurebrd @sweng66 ?

agr_literature_service/api/routers/topic_entity_tag_router.py
valearna: Good to know, it should be ok but I'll ask Ceri at standup.

agr_literature_service/api/crud/topic_entity_tag_crud.py
azurebrd: I'll defer to you both

valearna:
@sweng66 I modified the code to commit every 200 tags. How is it looking now?

agr_literature_service/api/routers/topic_entity_tag_router.py
valearna: Ceri approved this

azurebrd (approved):
Thanks @valearna

PR #591, opened 2024-01-31
adding code to handle duplicate tag rows with note info (merge note or force insertion)

sweng66:


azurebrd (approved):
Thanks @sweng66 .  I didn't look carefully, it looked like a bunch of code moved into a function.

sweng66:
yes, added code to take care of the duplicate tags with a note. also refactored the "create_tag" function to move the duplicate tag check code to another function to make code cleaner and more readable. Thanks for looking over! @azurebrd

PR #593, opened 2024-02-02
Use new session when querying db from separate process

valearna:
When querying the db from a subprocess of the api a separate session must be used. This PR adds code to create this session in the function to revalidate all tags


sweng66 (approved):
The fix Looks good ! Thanks @valearna

valearna:
Thanks @sweng66 !

PR #594, opened 2024-02-03
removed the script that validates DOI against PubMed

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #595, opened 2024-02-05
Scrum 3656 - TET entity_type must be null if entity is null

valearna:


sweng66 (approved):
Looks great! Thanks @valearna

PR #600, opened 2024-02-15
Fixed PubMed update program to FULLY update the data in the database based on the data from PubMed

sweng66:
plus sending a summary to slack report about the major changes (eg, journal, authors, volume, & issue) to the published papers.


azurebrd (approved):
thanks @sweng66 !

PR #602, opened 2024-02-15
fixed dqm loading script to work with newly updated 'update_authors' function

sweng66:
and add check before inserting a new PMCID into database (check if this ID is in database in invalid form, just in case)
also check if the paper has a resource info before accessing the resource title


azurebrd (approved):
Thanks @sweng66

PR #603, opened 2024-02-15
fixed to add the change the db connect

sweng66:


azurebrd (approved):
Ah, neat.  I kind of wondered about that, but figured that changing the x object would let it get added later on.  I guess you don't need a commit, because it's just adding things to the session, and the commit happens later ?

sweng66:
@azurebrd yes, pool all the changes and do a batch commit (250 papers at a time). 

Thank you so much for going over so many changes and so many PRs!!!

PR #604, opened 2024-02-15
fixed the code to not add a new paper if this paper doesn't have any …

sweng66:
…valid XREF IDs


azurebrd (approved):
Thanks @sweng66

sweng66:
Thanks @azurebrd

PR #608, opened 2024-02-28
Update the Ref DQM loading program to only send report to slack channel once a week (saturday)

sweng66:
but will send a report to slack channel if there is any error during data loading.


azurebrd (approved):
Thansk @sweng66 !

sweng66:
@azurebrd Thanks

PR #611, opened 2024-03-01
lookup constructs by modEntityId

ianlongden:
Again super slow due to very slow Ateam lookup. Ateam returns a load of stuff which could be why it takes so long?!!!! No idea but maybe eventually they will sort something out!


azurebrd (approved):
wow, maxret=2   If that's all it can handle, okay.  Thanks Ian.

ianlongden:
ah, the maxret was for testing. Put back to 1000. Good spot.

azurebrd (approved):
Thanks @ianlongden !

PR #617, opened 2024-03-11
adding code to retreive id to name mapping from a-team for affected genome model and it's subtypes (genotype, strain, & fish)

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #618, opened 2024-03-12
Added commands to restart api, automated scripts and debezium after applying alembic

valearna:


sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #622, opened 2024-03-14
Scrum 3662

ianlongden:


agr_literature_service/api/crud/utils/yml/reference.yml
```diff
@@ -0,0 +1,35 @@
+#
+#  NOTE: species has been left in here so that particular curie prefixes can be easily found
```
azurebrd, line 2: are species still here ?

agr_literature_service/api/crud/utils/yml/resource.yml
```diff
@@ -0,0 +1,39 @@
+#
+# This is pretty much the same as the reference apart from the ZFIN one.
+# It might be less duplication in the yml to copy the reference dict in the api
+# and then just make changes based on this file and just have difference ones here
```
azurebrd, line 4: I think only ZFIN and FB have resources, so it might be confusing if people see this and think it means there are resources for other mods ?  Might want some feedback from other developers, or even curators about future resource plans

agr_literature_service/api/routers/resource_router.py
```diff
@@ -69,3 +70,20 @@ def show(curie: str,
...
+
+
+@router.get('/check/patterns',
+            status_code=200,
+            )
+def show_resource_patterns():
+    return patterns_check.get_patterns()['resource']
+
+
+@router.get('/check/{curie_prefix}/{curie}',
```
azurebrd, line 82: This might be a little ambiguous, because it's expecting curie as a second parameter, but it's expecting to get the after-colon part of the curie instead of the whole curie.  Maybe check_pattern in patterns_check could do the check against the right side like now, but also against the curie_prefix:right_side passed in here, and return true if either passes ?  I'm not really sure.  Maybe it should expect the full curie, or it should explicitly ask for curie_right_side (don't know what to call that).  Would also welcome other dev feedback on this.

azurebrd (commented):
Looks good.  Had some thoughts in single comments, that would be nice to get some dev (or curator) feedback on, otherwise would approve.  Thanks.

agr_literature_service/api/routers/resource_router.py
ianlongden: My reasoning was that other code has to split this up anyway to store the prefix, so the splitting will have been done already. As for the ui I presumed less typing for curators was better so no point selecting MGI as the prefix and then typing in MGI:123456. It is easy to change but as you say it's up to the curators really. Happy either way.

agr_literature_service/api/crud/utils/yml/resource.yml
ianlongden: Okay that is easy, I can delete the other lines that are not needed. Thanks.

agr_literature_service/api/crud/utils/yml/reference.yml
ianlongden: Not species as such but at one point I thought only FB curators would be only able to add FB cross references and so we needed to have a way to determine who was allowed to add what (I thought FB curator would not be able to add an MGI cross ref, but people seemed happy with that). I left it in as that could change. Again its a simple change if people strongly feel it makes it confusing.

valearna (approved):

PR #624, opened 2024-03-16
Changed validation value for self validated tags

valearna:
Now using validated_right_self instead of self_validated


sweng66 (approved):
Looks good! Thanks @valearna

valearna:
Thanks @sweng66 !

PR #630, opened 2024-03-26
fixed to the image related extension list and removed the name has to matc…

sweng66:
and removed the name has to match 'fig' requirement when classifying 'figure' files


azurebrd (approved):
Thanks @sweng66 .  Why are some extensions kept in the comment, but not in the new code ?

sweng66:
@azurebrd Ceri suggested to not include some of the extensions. I leave those in the code just in case later we change our mind.
Thanks!

azurebrd:
Cool, thanks @sweng66 :)

PR #631, opened 2024-04-02
Fixed the python modules to check XREF IDs to make sure their formats are right before inserting them into DB

sweng66:
also fixed bug/typos in the PMID and RGD ID patterns  (in the reference.yml file)


azurebrd (approved):
Great, thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #632, opened 2024-04-02
Scrum 3728 rebuild search index cronjob

FB-ping:
it works fine on my local system with those steps to test:
1. start api/debezium/ui
2. check total numbers of records from search
3. delete a few records from elasticsearch index, recheck #2 see total numbers decrease.
4. start  automated_scripts, make it to start the setup crobtab
5.  after #4 done, go back #2 step and find that total number back to original.


valearna (approved):
Everything looks good now, thanks @FB-ping !

PR #633, opened 2024-04-03
fixed so it will set file class to 'thumbnail' if 'thumb' in the file name and file extension is in image related file extension..

sweng66:
otherwise will set the file class to 'figure' if it has one of image related file extensions


azurebrd (approved):
Thanks @sweng66

sweng66:
Thanks @azurebrd

PR #636, opened 2024-04-05
Scrum 3603

valearna:
Added list of ids of users who created tags that validate the current tag


pjhale (approved):
This looks fine but I'm still unable to test it.  If it does what its supposed to I think it will work fine.

PR #641, opened 2024-04-08
fixed to handle new parameter force_insertion

sweng66:


azurebrd (approved):
dunno how it works, but cool @sweng66 :)

sweng66:
Thanks @azurebrd

PR #650, opened 2024-04-25
adding/fixing code to work with nested data for topic_entity_tags and single tag (topic + confidence level) search

sweng66:
1. adding a new parameter 'tet_nested_facets_values' just for topic/confidence level selections
2. introducing a new aggregation branch to adjust the ES query to aggregate confidence levels within a specified topic,
    and to aggregate the topics within a specified confidence level to solve the counting issue when applying selections to 
    single tag
3. adding new function to deal with query for nested object (topic_entity_tags) 
4. processing search result to better handle and format the aggregation result for topic and confidence level data
5. update "add_curie_to_name_values" function to map topic curie to name correctly


valearna (approved):
Thanks @sweng66 ! We are currently counting the number of tags in the facets. It would be nice in the future to count the number of references and also to dynamically add facets to the queries instead of hardcoding their labels, but for this iteration what you have done is great!

sweng66:
Thanks @valearna

PR #652, opened 2024-04-27
fix those reference withotu citation_id on stage database

FB-ping:
all trigger scripts look fine, and data look OK in production, problem only occured in stage database, so we suspect that for some reason, the initial step to generate reference.citation for all records were interrupt. so we only need to run script to fix those without ciation_id on stage database. 

docker exec -it agr-literature-rdsdev-4001_api_1 sh
 python3 agr_literature_service/lit_processing/oneoff_scripts/add_citation.py


sweng66 (approved):
Looks good! Thanks @FB-ping

PR #659, opened 2024-05-09
Scrum 1999 - API changes to merge TET data

valearna:
- Merge TET entries during reference merge API call
- Revalidate tags of winning paper to account for possible cross-reference validation between tags from losing and winning refs
- Fixed date_updated management for note updates on duplicate tags with different note values
- Changed force_insertion behavior for TET API - now only forces insertion on different created_by users and not for the very same tag by the same user

### DIFF
--- a/agr_literature_service/api/crud/reference_crud.py
+++ b/agr_literature_service/api/crud/reference_crud.py
@@ -33,9 +33,10 @@ from agr_literature_service.api.models import (AuthorModel, CrossReferenceModel,
                                                ReferenceModel,
                                                ResourceModel,
                                                CopyrightLicenseModel,
-                                               CitationModel)
+                                               CitationModel, TopicEntityTagModel)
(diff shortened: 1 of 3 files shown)


sweng66 (approved):
Looks great! Thanks! @valearna

valearna:
I have not yet tested this with a large number of TETs. I'll keep the PR open and add a specific webtest with a large number of tags

valearna:
Thanks for the review @sweng66

valearna:
Added a webtest for large merges. It took ~2m to merge 2 refs with 1k tets each. The time is used mostly to check for duplicates. I will try to optimize the create_tag function to see if we can get a bit faster

azurebrd (commented):
Thanks @valearna !

valearna:
Thanks @azurebrd @sweng66. This is ready for a final review

sweng66 (approved):

PR #661, opened 2024-05-13
Improved queries to speed up validation

valearna:
using subset of fields in the overall query and raw sql to insert validation ids


valearna:
merge time for 2 refs with 1k tags each down to 45s from 1m 15s.

sweng66 (approved):
Although I am not sure I understand the details, but It looks good ! Thanks @valearna

valearna:
thanks @sweng66 !

PR #662, opened 2024-05-20
adding code to validate/map ID to name for complex and pathway against SGD API

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #663, opened 2024-05-21
Scrum 3678 facet for TET source fields

valearna:
- Refactored code to make it easier to add new TET fields and to avoid extra API calls
- Added aggregations for source fields
- Added code to convert eco curies to names


sweng66 (approved):
Looks great! Thanks @valearna

valearna:
Thanks @sweng66 !

PR #669, opened 2024-06-05
refactoring the 'update_authors' function to try to update author rows as much as possible

sweng66:
* Perform case-insensitive comparisons of author names after trimming leading and  trailing whitespace
* Update existing author records to maintain author IDs as much as possible by comparing the key (combination of name, last_name, first_name, & first_initial), orcid ID, combination of lastName and firstInitial.
* Handle updates to affiliations and ORCIDs without deleting the entire author record
* Update author names if the order remains the same to maintain author IDs
* Update author names if the order is changed by using temp order mapping

### DIFF
--- a/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/pubmed_update_references_single_mod.py
+++ b/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/pubmed_update_references_single_mod.py
@@ -76,7 +76,7 @@ def update_data(mod, pmids, resourceUpdated=None):  # noqa: C901 pragma: no cove
         email_subject = mod + " " + email_subject
 
     # remove old log files
-    remove_old_files(log_path, 30)
+    remove_old_files(log_path, 90)
 
     # set new log file with date stamp
     datestamp = str(date.today()).replace("-", "")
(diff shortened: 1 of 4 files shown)


azurebrd (approved):
Wow, it's so complex, so many different cases and exceptions, that's amazing, @sweng66 !

I'm not sure that we can really know some of the things that are being asserted, I mean, with absolute certainty, but it seems close enough.  In general people don't complain if stuff is wrong at WB, so if authors get reassigned to authors automatically, and it's usually correct, the few cases where PubMed messed it up and makes it be a wrong reassignment because Li H and H Li were actually different authors in the same paper, well, actually, I think you account for that also, that it has to match exactly to one thing.  So it's probably all great, thank you Shuai !

sweng66:
@azurebrd Thanks for checking this out.  Yes, it is complicated. 

The end result from the author update should exactly match what PubMed has so I think it should be okay.

We will not update entries flagged as first_author, corresponding_author, or with a PERSON connection unless we are certain about the changes. This can be redefined later when the process is ready.

I have tested all PubMed updates over the past five weeks for all papers in ABC to make sure all cases are caught.

azurebrd:
That's awesome, thanks @sweng66 !

azurebrd (approved):
Nice !

PR #670, opened 2024-06-10
Refactored the update_authors function by creating and utilizing an Author class that encapsulates functionality related to author keys and comparisons

sweng66:


valearna (approved):
Thanks @sweng66 !

PR #673, opened 2024-06-14
Scrum 4112

ianlongden:


ianlongden:
I think I got the transitions correct but they may need checking.
So alembic will need running an update before the script to add data can be applied.
Available via slack if needed but I will be away from my laptop now.

valearna (approved):
Thanks @ianlongden ! We have changed transitions for the file upload workflow, but we can change them later in the db

PR #675, opened 2024-06-22
Scrum 4117 - Functions and API endpoints to manage workflow transitions

valearna:


azurebrd (dismissed):
Thanks @valearna .  Not sure I really get it, but the code makes sense.  If someone else understands better how this should work, it might be better for them to review it, or maybe go over it together as a group, so we all get how it should work.  Unless this is still a test to see if this will work for curators and there could be multiple iterations like topic_entity_tag, in which case let's wait until it's stable.  Thanks !

valearna:
Thanks @azurebrd ! Yes, I prefer to talk about this to the group next week before merging to make sure the new functionality makes sense to everybody.

sweng66:
Everything is looking great. Thanks @valearna 

but I do have a question about the following function:

https://github.com/alliance-genome/agr_literature_service/blob/SCRUM-4117/agr_literature_service/api/crud/workflow_tag_crud.py#L39

def get_parent_or_children(atp_name: str, parent_or_children: str = "parent"):
    workflow_children, workflow_parent = load_workflow_parent_children()
    workflow_to_check = workflow_children if parent_or_children == "children" else workflow_parent
    if atp_name not in workflow_to_check:
        logger.error("Could not find parent for {}".format(atp_name))
        return None
    return workflow_to_check[atp_name]

Should we change "logger.error("Could not find parent for {}".format(atp_name))" to the following?

 if atp_name not in workflow_to_check:
        logger.error("Could not find parent or children for {}".format(atp_name))
        return None

Also where do we load the workflow parent/children relationship and where is the "load_workflow_parent_children()" defined in the code? we may have talked about this in one of the backlog refinements, but for some reason I missed this part.

Anyway, we can talk about these in next Monday's standup. Thanks!

sweng66 (dismissed):

sweng66:
nevermind. Saw the function "load_workflow_parent_children" @valearna

valearna:
Thanks @sweng66 ! I've updated the error message to show "parent" or "children" depending on the parent_or_children parameter

sweng66 (approved):
Great! Thanks @valearna

PR #678, opened 2024-06-27
Scrum 4170 - validating entities based on entity_id_validation

valearna:
* added WB validation
* refactored code to avoid duplications


sweng66 (dismissed):
Looks good! Thanks @valearna

azurebrd (commented):
@sweng66 @valearna Do we need to put SGD_API_URL and WB_API_URL in the .env  so we know that it needs to be added in the filesystem version ?

topic_entity_tag_crud.py has a commented out import in line 36 we probably don't need, but it's fine

sweng66:
@azurebrd both suggestions (adding WB/SGD URLs to .env and removing the line36 from tet crud) sound good to me.

valearna:
Yes, thanks @azurebrd ! I'll add them to the .env file

azurebrd (approved):
Thanks for the changes @valearna !

valearna:
Thanks @azurebrd and @sweng66

PR #679, opened 2024-06-27
fixed the sorting issue for the papers with no publication date

sweng66:


valearna (approved):
Thanks @sweng66 ! Using dict.get() instead of the ifs makes the code much more readable!

sweng66:
Thanks! @valearna

PR #682, opened 2024-07-01
adding a new endpoint for exporting tet data for a given mod

sweng66:


agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -586,6 +587,46 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1,
...
+    current_date = datetime.now()
+    past_date = current_date - timedelta(days=int(days_updated))
+    last_date_updated = past_date.strftime("%Y-%m-%d")
+
+    rows = db.execute(f"SELECT cr.curie, tet.*, u.email "
+                      f"FROM cross_reference cr "
+                      f"JOIN topic_entity_tag tet ON cr.reference_id = tet.reference_id AND cr.curie_prefix = '{mod_abbreviation}' "
+                      f"JOIN topic_entity_tag_source tets ON tet.topic_entity_tag_source_id = tets.topic_entity_tag_source_id "
+                      f"JOIN users u ON tet.updated_by = u.id "
+                      f"WHERE tets.data_provider = '{mod_abbreviation}' "
```
valearna, line 601: I'm not sure if we should filter by data_provider or secondary_data_provider here

agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -586,6 +587,46 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1,
         return [get_tet_with_names(db, tag, curie_to_name) for tag in all_tet]
 
 
+def get_reference_tags(db: Session, mod_abbreviation: str, days_updated: int = 7):
```
valearna, line 590: Can we call the function `get_all_topic_entity_tags_by_mod` or something similar to make it clear that it's not by reference but for all tags in all references for the mod?

agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -586,6 +587,46 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1,
...
+
+    data = [get_tet_with_names(db, tag, curie_to_name_mapping) for tag in tags]
+
+    src_rows = db.execute(f"SELECT * FROM topic_entity_tag_source WHERE data_provider = '{mod_abbreviation}'").fetchall()
+    metadata = [dict(row) for row in src_rows]
+
+    return {"metadata": metadata, "data": data}
+
+
+def get_curie_to_name_mapping_for_mod(db, mod_abbreviation, last_date_updated=None):
```
valearna, line 616: maybe we could merge this function and `get_curie_to_name_from_all_tets` and do it for all references if a reference_id is not passed?

agr_literature_service/api/routers/topic_entity_tag_router.py
```diff
@@ -132,6 +132,14 @@ def show_all_reference_tags(curie_or_reference_id: str,
                                                          column_values)
 
 
+@router.get('/by_mod/{mod_abbreviation}',
+            status_code=200)
+def get_reference_tags(mod_abbreviation: str,
```
valearna, line 137: Same as above, I would rename this endpoint to make it clear that this is for all tets for all references for a mod

valearna (changes requested):

agr_literature_service/api/crud/topic_entity_tag_crud.py
sweng66: We have "secondary_data_provider_id" column in tet source table, but there is nothing in this column, but "data_provider" column contains mod abbreviations.

valearna: That's the mod_id and we should use that column. I don't know why it's empty but it shouldn't be. data_provider is the original data provider, which can be something different than a mod_abbreviation if the data was originally generated by an external resource

agr_literature_service/api/crud/topic_entity_tag_crud.py
sweng66: Sounds good!

agr_literature_service/api/crud/topic_entity_tag_crud.py
sweng66: Did you mean that we can update the "get_curie_to_name_from_all_tets" function to take optional parameters "mod_abbreviation", "reference_curie_or_id", "date_last_updated". If none of these are provided, getting mapping for all papers in tet table? We can do this, but I will just leave as is since the "get_curie_to_name_from_all_tets" function is pretty complicated now.  What do you think?

agr_literature_service/api/routers/topic_entity_tag_router.py
sweng66: Fixed =)

agr_literature_service/api/crud/topic_entity_tag_crud.py
sweng66: fixed =)

sweng66:
@valearna Thanks for looking over the code. I updated the function name to "get_all_topic_entity_tags_by_mod" (good suggestion!), but leave the "get_curie_to_name_mapping_for_mod" function as is since this maybe a little cleaner?

azurebrd (commented):
This looks fine to me, but it seems @sweng66 and @valearna are working out some fine details, so I'm just leaving a comment instead of approval and merge, in case there's still that one thing to resolve.  Thanks

valearna:
The code looks good now, thanks @sweng66. I still see data_provider instead of secondary_data_provider. secondary_data_provider is the mod and data_provider is not necessarily a mod, so I think we should really use secondary_data_provider

valearna:
I checked both stage and prod and it looks like all sources have a non null secondary_data_provider_id. That field is nullable=False. It's a foreign key on mod.mod_id, so you'll need to add mod to the join or read the mod_id from the mod table in a separate query

valearna:
I find the naming data_provider and secondary_data_provider very confusing, but we are using the definitions of these fields in the LinkML model, so hopefully the curators will understand what these fields are

sweng66:
Good catch! I totally missed the mod_id in the secondary_data_provider_id lol. @valearna 

I will fix the code to use this column. 

Thanks @azurebrd for checking out the code!

sweng66:
@valearna Fixed to join the sql with secondary_data_provider_id (mod_id). Thanks.

agr_literature_service/api/crud/topic_entity_tag_crud.py
```diff
@@ -586,6 +587,50 @@ def show_all_reference_tags(db: Session, curie_or_reference_id, page: int = 1,
...
+
+
+def get_curie_to_name_mapping_for_mod(db, mod_abbreviation, last_date_updated):
+
+    curie_to_name_mapping = {}
+
+    rows = db.execute(f"SELECT DISTINCT tet.reference_id "
+                      f"FROM topic_entity_tag tet "
+                      f"JOIN topic_entity_tag_source tets ON tet.topic_entity_tag_source_id = tets.topic_entity_tag_source_id "
+                      f"WHERE tets.data_provider = '{mod_abbreviation}' "
```
valearna, line 627: Use secondary_deta_provider also here

sweng66: oops will fix that.

sweng66:
@valearna fixed. Thanks!

valearna (approved):
Thanks @sweng66 !

sweng66:
Thanks for the great code review! @valearna

PR #683, opened 2024-07-05
create new workflow tag object if it does not exist

valearna:


azurebrd (approved):
Thanks @valearna !

valearna:
Thanks @azurebrd !

PR #687, opened 2024-07-11
fixed to move all papers that do not have a valid MOD curie out of MOD corpus

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #693, opened 2024-07-18
Adding Language

pjhale:
Added language to the API returns.  Also added char filter to change all options to either English or Other.

Needs UI PR to work.


valearna (approved):
Thanks @pjhale !

PR #694, opened 2024-07-19
Scrum 4028 speed up revalidation of topic entity tags

valearna:
* caching list of related tags for the same reference and mod
* removed extra commits
* added function to avoid automatically setting updated_by to api user when revalidating tags
* removed validation values from tags comparison to find duplicates (these values are set after the tag is inserted)
* used join to speed up query to retrieve all tags and their related info on subobjects


valearna:
I'm currently running the function to revalidate all tags and calculate the validation values to be stored in the new db field (instead of calculating them on the fly as we used to do). It still takes hours but hopefully not days as before.

azurebrd (approved):
I don't really follow it, but approving.  If you want someone that understands it to review it, that's better

valearna:
Thanks @azurebrd !

PR #701, opened 2024-07-26
fixed the dqm loading script and related functions to store WB three referencetypes/tags in MRT table instead of WFT table

sweng66:


azurebrd (approved):
Great, thanks @sweng66 !  It looks like if there's a conflict, it will always prefer Meeting_abstract .  I think it's fine, since curators will get a report.  (as opposed to having to ask Kimberly if she'd rather have a priority structure)

sweng66:
yes. Thanks @azurebrd

PR #703, opened 2024-07-31
fixed textpresso reference retrieval API to limit WB papers for certain species

sweng66:


valearna (approved):
Thanks @sweng66. Can you let Michael know that we modified the API parameters when we deploy to prod?

sweng66:
Thanks @valearna 
I will let Michael.

PR #704, opened 2024-08-01
Scrum 4310 fixed validation bugs

valearna:
- automatic update of date_updated now disabled when setting validation values
- calculating validation values for all types of created tags (even those that are not validating other tags)


azurebrd (approved):
Thanks @valearna !

PR #708, opened 2024-08-12
Transition data

ianlongden:


valearna:
@ianlongden shouldn't the table to file script read also the new columns (actions and condition)?

ianlongden:
They should now but these were only just added to stage and were not there when the script was made/used. So would have crashed earlier on. But I can fix this now.

valearna:
Thanks @ianlongden !

valearna (dismissed):

valearna (approved):

PR #710, opened 2024-08-12
adding code for generating SGDID when a new paper is added to SGD mod…

sweng66:
… corpus through ABC


sweng66:
Thanks @valearna

PR #714, opened 2024-08-15
new script for dumping new SGD references with new sgdids 

sweng66:


crontab
```diff
@@ -1,9 +1,10 @@
-SHELL=/bin/bash
+cSHELL=/bin/bash
```
valearna, line 1: Is this an intended change or a typo? If it's intended, what does it do?

valearna (dismissed):
Looks good to me. Thanks @sweng66

crontab
sweng66: oops good catch! will fix

sweng66: fixed

sweng66:
Thanks @valearna

sweng66:
@valearna can you reapprove this pr? thanks

valearna (approved):

PR #736, opened 2024-08-26
Remove wf tag if mca is deleted or corpus is set to != true + add it if no other tags are present

valearna:


valearna:
@azurebrd @sweng66 I've tested this PR by creating a new reference manually, setting it outside_corpus, and then inside_corpus again. I've also tried setting it to needs_review and different combinations of the possible states. Please make sure that this covers everything we discussed before. The function that generates MOD ids already checks if one exists (a non-obsolete one) and doesn't create a new MOD ID if not needed.

azurebrd (approved):
Great, thanks @valearna ! 

I haven't looked at a lot of the API code for WF, but I see that ATP 141 is hardcoded in.  It seems generally bad that the ontology can change somewhere without us knowing, but the code is hardcoded to expect something.  I can't think of another way around it.  But I'm wondering if what we want is to have the ontology hardcoded into the repo instead.  Then we don't have to look up anything, and when a curator wants to change the ontology, they should edit the version in our repo, and make some kind of check with a developer that they're not breaking anything against hardcoded values.

I get that in theory, "they're not going to change the ontology".  So I don't get why we should have an ontology, instead of just hardcoding stuff in our system, or is someone else meant to use ATP besides Blue Team ?  I can see that if other people are using it, someday someone will change it without knowing how much stuff they're breaking here.

valearna:
Thanks @azurebrd. I think that it's ok to use ATP IDs, at least for this special case where 141 is the starting point for all workflows. The ontology structure and the labels may change but that node ID should remain the same.

PR #738, opened 2024-08-27
New API endpoint to fetch references with a specific WFT generated within a given period for a mod

sweng66:


sweng66:
Thanks @valearna 

Good suggestion!  We can use 'all' for all mods. This way, we can use same endpoint name. What do you think?

workflow_tag/by_mod/SGD?workflow_tag_id=ATP:0000141&startDate=2024-08-23
workflow_tag/by_mod/all?workflow_tag_id=ATP:0000141&startDate=2024-08-23

sweng66:
yes, but we may need a parameter for this, something like root_workflow_tag_id.  @ianlongden

PR #743, opened 2024-09-03
fixed to sort by date_published_start since it is type of date and we…

sweng66:
… dont need to use script and all


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #744, opened 2024-09-03
fixes

ianlongden:


ianlongden:
prod has had the changes to the transition table added. stage looked okay.

valearna (approved):
Thanks @ianlongden! Do we need to deploy anything to prod?

ianlongden:
At some point but no rush. It just gives a better error message if the method name is messed up.

PR #749, opened 2024-09-09
fixed to map the curies to themselves if any curies in the chunk were not returned by the API

sweng66:


valearna (approved):
Thanks @sweng66 !

sweng66:
Thanks @valearna

PR #751, opened 2024-09-11
adding fallback to grab entity id to name mapping from SGD

sweng66:


azurebrd (dismissed):
Thanks @sweng66 !

There's some commented out code on 334 335 342 343, not sure if you'd want a comment about what it would be used for later, or just remove, but it's fine to just leave in.

sweng66:
@azurebrd Good catch! I just removed these 4 commented old code. Thanks!

sweng66:
@azurebrd Can you check it again? Thanks

azurebrd (approved):
Cool, thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #763, opened 2024-10-04
fixing unit tests + adding 'with db.begin()' to raw sqls for INSERT/UPDATE/DELETE statements

sweng66:


azurebrd (approved):
Thanks @sweng66 !

PR #764, opened 2024-10-06
fixing mypy errors

sweng66:


azurebrd (approved):
Looks good, thanks @sweng66 !  That's so much stuff.  For the test_reference.py I don't think the tets len should be 1, but I think @valearna should look at it

PR #769, opened 2024-10-09
adding two new fields to schema file for TET endpoint

sweng66:


FB-ping (approved):

valearna:
We shouldn't merge this as the fields are already there on the main branch. We should merge main into 4458

PR #770, opened 2024-10-10
SQLAlchemy upgrade from version 1.2 to 2.0.7

sweng66:


agr_literature_service/api/crud/reference_crud.py
```diff
@@ -780,99 +784,138 @@
...
-                         FROM cross_reference
-                         WHERE curie_prefix='{curie_prefix}') as ref_mod
-               WHERE sub_select.reference_id=reference.reference_id
-               AND sub_select.reference_id=ref_pmid.reference_id
-               AND sub_select.reference_id=ref_doi.reference_id
-               AND sub_select.reference_id=ref_mod.reference_id
-               AND reference.citation_id=citation.citation_id
-               ORDER BY date_created {order_by}
-           """
+    query = text(query_str)
```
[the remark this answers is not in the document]
sweng66: fixed.

agr_literature_service/api/crud/reference_crud.py
```diff
@@ -889,102 +934,148 @@ def get_bib_info(db, curie, mod_abbreviation: str, return_format: str = 'txt'):
...
-      AND rf.file_class = 'main'
-      AND rf.file_extension = 'pdf'
-      AND (rfm.mod_id is NULL OR rfm.mod_id = {mod_id})"""
+    query_params = {'mod_id': mod_id}
 
+    # Add condition for reference_type if provided
     if reference_type:
-        select_stmt += f"""
+        query_str += """
         AND r.reference_id IN (
```
azurebrd, line 1015: Did sqlalchemy not like the old syntax and wanting JOIN, or was it something else ?  I recall it didn't like {mod_id} and wanting :mod_id  but does :mod_id not work without JOIN or something like that ?

agr_literature_service/api/models/mod_model.py
```diff
@@ -43,13 +43,16 @@ class ModModel(Base, AuditedModel):
...
+    taxon_ids: Column = Column(
         ARRAY(String()),
         unique=False,
         nullable=True
     )
 
-    referencetypes = relationship("ModReferencetypeAssociationModel")
+    referencetypes = relationship(
+        "ModReferencetypeAssociationModel",
+        back_populates="mod"
```
azurebrd, line 54: This is actually needed now, and it wasn't needed before ?  Just making sure because sometimes extra back_populates were making things small, and @valearna removed them

agr_literature_service/api/models/mod_reference_type_model.py
```diff
@@ -44,15 +44,15 @@ class ModReferencetypeAssociationModel(Base):
...
 
-    mod_id = Column(
+    mod_id: Column = Column(
         ForeignKey("mod.mod_id", ondelete="CASCADE"),
         index=True,
         nullable=False
     )
 
-    mod = relationship("ModModel")
+    mod = relationship("ModModel", back_populates="referencetypes")
```
azurebrd, line 53: same question as in mod_model.py

agr_literature_service/lit_processing/utils/db_read_utils.py
```diff
@@ -51,30 +51,29 @@ def get_curie_to_title_mapping(curie_list):
...
 
     curies = ", ".join(["'" + x + "'" for x in curie_list])
-    rs = db_session.execute("SELECT curie, title FROM reference WHERE curie IN (" + curies + ")")
+    rs = db_session.execute(text("SELECT curie, title FROM reference WHERE curie IN (" + curies + ")"))
     rows = rs.fetchall()
 
-    ref_curie_to_title = {}
-    for x in rows:
-        ref_curie_to_title[x[0]] = x[1]
+    ref_curie_to_title = {x[0]: x[1] for x in rows}
```
azurebrd, line 57: Oh, neat, that looks so Valerio

tests/api/test_mod_ref_type.py
```diff
@@ -105,21 +106,27 @@ def test_destroy_mrt(self, test_mod_ref_type, auth_headers): # noqa
             assert response.status_code == status.HTTP_404_NOT_FOUND
 
     def test_display_order(self, db, test_mod_ref_type, auth_headers): # noqa
-        mod_id = db.query(ModModel.mod_id).filter(ModModel.abbreviation == "ZFIN").one_or_none()
+        # mod_id = db.query(ModModel.mod_id).filter(ModModel.abbreviation == "ZFIN").one_or_none()
+        mod_id = db.query(ModModel.mod_id).filter(ModModel.abbreviation == "ZFIN").scalar()
```
azurebrd, line 110: In perl scalar(@array) would give you the length of the array.  Is this getting a scalar by extracting the first value, like .one() used to do ?  or is it context dependent ?

tests/api/test_reference.py
```diff
@@ -317,7 +321,7 @@ def test_reference_large(self, db, auth_headers, populate_test_mod_reference_typ
             assert response['mesh_terms'][0]['heading_term'] == "hterm"
 
             # cross references in the db?
-            xref = db.query(CrossReferenceModel).filter(CrossReferenceModel.curie == "FB:FBrf0221304").one()
+            xref = db.query(CrossReferenceModel).filter(CrossReferenceModel.curie == "FB:FBrf0221304").first()
```
azurebrd, line 324: Wondering with previous comment, because before it became scalar() and here it's becoming first()

tests/api/test_topic_entity_tag.py
```diff
@@ -128,8 +128,12 @@ def test_patch(self, test_topic_entity_tag, auth_headers): # noqa
                 "topic": "new_topic",
                 "entity_type": "new_type",
                 "entity": "new_entity",
-                "updated_by": "new_user",
-                "novel_topic_data": False,
+                "entity_id_validation": "alliance",
```
azurebrd, line 131: There's no additional testing for these fields in the response.  Does patch now require adding all fields ?  Before we'd use PUT for that, and we used PATCH because it didn't require everything

tests/lit_processing/data_ingest/dqm_ingest/utils/test_md5sum_utils.py
```diff
@@ -1,103 +1,176 @@
...
+        # Assertions to verify no changes
+        md5sum_results = db.execute(text("""
+            SELECT rmm.md5sum
+            FROM cross_reference r
+            JOIN reference_mod_md5sum rmm ON r.reference_id = rmm.reference_id
+            JOIN mod m ON rmm.mod_id = m.mod_id
+            WHERE m.abbreviation='FB' AND r.curie='FB:FBrf00000001'
+        """))
+        md5sums = md5sum_results.mappings().fetchall()
+        assert md5sums, "No md5sum found for FB:FBrf00000001 after empty update"
```
azurebrd, line 103: Does this assert that md5sums exists or isn't None or has some result ?

azurebrd (approved):
Thanks @sweng66 !  Looks great.  I had some questions as comments, but I'm sure it's all good.  That was so much, way to go !

agr_literature_service/api/crud/reference_crud.py
sweng66: Good question! @azurebrd  The old syntax is still working for version 2.0.7 of SQLAlchemy.
Just JOIN makes things clear (at least to me). The :mod_id should work both with and
without a JOIN. The {mod_id} syntax is still working in this version (2.0.7), but it may have
potential security concerns although in our case it should not be a big deal (since we are
using the query to select the data from our database). Anyway, github security check wasn't
happy about it and the SQLAlchemy 2.x does enforce stricter SQL parameterization rules, and
the :mod_id syntax meets the SQL standard of binding parameter requirement for the SQLAlchemy 2.x.

agr_literature_service/api/models/mod_model.py
sweng66: I added the back_populates="mod" in the referencetypes relationship to establish
a bi-directional relationship between the ModModel and ModReferencetypeAssociationModel classes.
This allows SQLAlchemy to automatically manage the relationship in both directions.

So when we access mod.referencetypes (on ModModel), it fetches the related
ModReferencetypeAssociationModel instances associated with that mod.
When we access referencetype.mod (on ModReferencetypeAssociationModel), it fetches the
associated ModModel instance.

Without adding this, all programs and modules using mod generate a BIG warning. Remember we
have talked about this in slack channel. And I asked ChatGPT for this solution.

agr_literature_service/api/models/mod_reference_type_model.py
sweng66: same reason as previous one

agr_literature_service/lit_processing/utils/db_read_utils.py
sweng66: LOL. Indeed.

tests/api/test_mod_ref_type.py
sweng66: In SQLAlchemy, the method .scalar() is used to return the first column of the first row of the query result, or None if no rows are found. It is not equivalent to Perl's scalar(@array) which gives the length of an array. The .scalar() in SQLAlchemy is used for when you expect a single value from the query and do not need the full row object.

For example,  mod_id = db.query(ModModel.mod_id).filter(ModModel.abbreviation == "ZFIN").scalar()
we only query **ModModel.mod_id**

   we can rewrite it to :

   mod = db.query(ModModel).filter(ModModel.abbreviation == "ZFIN").first()
   mod_id = mod.mod_id
   
   Does this make sense?

tests/api/test_reference.py
sweng66: The method .first() fetches the first row of the result set and returns it as an object since we are querying CrossReferenceModel instead of CrossReferenceModel.curie.

   we can also rewrite it to:

curie = db.query(CrossReferenceModel.curie).filter(CrossReferenceModel.curie == "FB:FBrf0221304").scalar()

tests/api/test_topic_entity_tag.py
sweng66: Good catch! We don't need that, but I leave as is for now since it will not affect the test (too lazy to change) =)

tests/lit_processing/data_ingest/dqm_ingest/utils/test_md5sum_utils.py
sweng66: Yes, this assert statement checks if md5sums exists, or it is not None, not empty, or not equivalent to False

sweng66:
Many thanks @azurebrd for carefully looking over the PR !!

agr_literature_service/api/crud/reference_crud.py
azurebrd: Cool cool, good to know, thanks !

agr_literature_service/api/models/mod_model.py
azurebrd: Ah, okay.  I saw that you and Valerio were discussing it, but I wasn't clear.

tests/api/test_mod_ref_type.py
azurebrd: Yes, thanks !

tests/api/test_topic_entity_tag.py
azurebrd: Cool cool :)

tests/lit_processing/data_ingest/dqm_ingest/utils/test_md5sum_utils.py
azurebrd: cool cool

tests/api/test_reference.py
azurebrd: got it, thanks !

PR #779, opened 2024-11-01
sql_query_for_missing_files was not working on just missing files, bu…

azurebrd:
…t workflow tag status, renamed function.  subquery LEFT JOIN referencefile was making it very slow, instead processing the reference list subset to display, and query just that set for referencefile information, then aggregate that into the resulting ref_data to return.

Also, the default query searches for either of two specific ATP terms, instead of the exclusion of two other specific ATP terms

### DIFF
--- a/agr_literature_service/api/crud/reference_crud.py
+++ b/agr_literature_service/api/crud/reference_crud.py
@@ -786,52 +786,37 @@ def add_license(db: Session, curie: str, license: str):  # noqa
 
 def sql_query_for_workflow_files(db: Session, mod_abbreviation: str, order_by: str, filter: str, offset: Optional[int] = None, limit: Optional[int] = None):
 
-    subquery: Optional[TextClause] = None
     curie_prefix = 'Xenbase' if mod_abbreviation == 'XB' else mod_abbreviation
+    workflow_tag_id_clause: Optional[TextClause] = None
 
     if filter == 'default':
-        subquery = text("""
-            SELECT b.reference_id
-            FROM mod_corpus_association AS b
-            JOIN mod ON b.mod_id = mod.mod_id
-            LEFT JOIN workflow_tag AS d ON b.reference_id = d.reference_id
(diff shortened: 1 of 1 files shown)


agr_literature_service/api/crud/reference_crud.py
```diff
@@ -864,7 +854,36 @@ def sql_query_for_missing_files(db: Session, mod_abbreviation: str, order_by: st
...
         params['offset'] = str(offset)
 
-    return query.bindparams(**params)
+    rows = db.execute(query, params).mappings().fetchall()
+    ref_data = jsonable_encoder(rows)
+    if not ref_data:
+        return ref_data
+
+    reference_ids = [item['reference_id'] for item in ref_data if 'reference_id' in item]
+    reference_ids_sql_txt = ', '.join(str(item) for item in reference_ids)
```
sweng66, line 863: Can we  eliminate the intermediate reference_ids_sql_txt and directly use bind parameter as follows? 

reference_ids = [item['reference_id'] for item in ref_data if 'reference_id' in item]
reffile_query_str = """
    SELECT reference_id,
           COUNT(1) FILTER (WHERE file_class = 'main') AS maincount,
           COUNT(1) FILTER (WHERE file_class = 'supplement') AS supcount
    FROM referencefile
    WHERE reference_id = ANY(:reference_ids)
    GROUP BY reference_id
"""

rows_reffile = db.execute(text(reffile_query_str), {'reference_ids': reference_ids}).mappings().fetchall()

sweng66 (approved):
Looks great! Thanks @azurebrd

agr_literature_service/api/crud/reference_crud.py
azurebrd: Oh, cool !  I didn't know the syntax for that, that looks much better, I'll try it after stand up, thanks !

sweng66: The fix looks great. Thanks

sweng66 (approved):

azurebrd:
Thanks @sweng66 !

PR #782, opened 2024-11-05
added memory constraints to jenkins

valearna:


valearna:
This is already running on the dev server and there is more available memory now @FB-ping.

sweng66 (approved):
Looks good. Thanks @valearna

valearna:
Thanks @sweng66 !

PR #783, opened 2024-11-06
adding a new function for updating the md5sum correctly 

sweng66:
Also updated the dqm loading program to call this new function


valearna (approved):

sweng66:
Thanks @valearna

PR #784, opened 2024-11-07
a new script to set the pubmed_types to NULL if a paper has an invalid PMID 

sweng66:
and will remove a paper if this paper is not associated with any mods and doesn't have a PMID, also this paper has a duplicate title with another paper in the database.


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #787, opened 2024-11-14
Scrum 4360 workflow tag search

FB-ping:
for workflow_tag facet


sweng66 (approved):
Looks good to me. Thanks @FB-ping

PR #793, opened 2024-11-26
updating api endpoint to deal with WFT change when changing a file status 

sweng66:
This will work only after a new transition ("files uploaded" to "file upload in progress") is added into WF transition table.


valearna (approved):

sweng66:
Thanks @valearna

PR #795, opened 2024-12-02
Scrum 4579 - do not set the file conversion to "failed" in case of network errors

valearna:


sweng66:
Do we not want to set the job in progress or failed during the pdf to text conversion? when do we set the status? @valearna

valearna:
@sweng66 at standup we decided that we won't add "in progress" for pdf2tei conversion and we don't need to change the status to "failed" in case of network errors so that we can try again at the next run

sweng66 (approved):
Looks good! Thanks @valearna

valearna:
thanks @sweng66 !

PR #797, opened 2024-12-05
transition table fixes

ianlongden:


sweng66 (commented):
Looks good to me, but will defer to Valerio. Thanks @ianlongden

PR #798, opened 2024-12-05
Updated the code to handle multiple curators simultaneously generating MOD CURIEs (e.g., WB Paper ID or SGDID)

sweng66:
Also updated code to return meaningful warning message to let curator know that another curator may work on the same paper at the moment.


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #799, opened 2024-12-06
check if title is extracted, if not set failure 

valearna:
I found some special cases where grobid generates a valid tei file but this is empty


sweng66 (dismissed):
Looks good. thanks @valearna

sweng66 (approved):

PR #801, opened 2024-12-11
Added custom jsonable encoder to avoid getting trapped into circular …

valearna:
…relationships


sweng66 (approved):
Cool. Thanks @valearna

PR #802, opened 2024-12-12
adding a new script for checking and setting correct file upload workflow tags 

sweng66:


azurebrd (approved):
I don't really get the WF stuff, but since it's a one off, if it works, cool, thanks @sweng66 .

sweng66:
Thanks @azurebrd

PR #804, opened 2024-12-12
refreshing dqm loading logs even though there is nothing to log

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #806, opened 2024-12-13
end date needs to add a date, or it will treat it as the start of the…

azurebrd:
… day, but treating it as a timestamp didn't work somehow, so converting back to string


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #808, opened 2024-12-14
Valerio and Ceri modified this so that it always requires it be insid…

azurebrd:
…e corpus, so needed to update the query to always join on reference and mca


sweng66 (approved):

azurebrd:
Thanks @sweng66 !

PR #810, opened 2024-12-16
adding code to clean up text conversion / ref classification / entity…

sweng66:
… extraction WFTs for not a real main PDF


valearna (approved):

sweng66:
Thanks @valearna

PR #813, opened 2025-01-07
updated annual PubMed full update script to call functions in the main script

sweng66:
plus an one-off script to obsolete a list of PubMed IDs.


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #814, opened 2025-01-10
Scrum 4736 - process all pdfs that need to be converted when running the pdf2tei converter instead of just the first 1k

valearna:


valearna:
Thanks @sweng66 !

valearna:
This change will only affect prod

PR #815, opened 2025-01-15
SCRUM-4609 return ref curie in jobs

ianlongden:


valearna (approved):
Thanks @ianlongden! I'll update the classification pipeline to use the new field too.

PR #816, opened 2025-01-22
Fixed query to get list of jobs + added more checks and logs to pdf2tei script

valearna:


valearna:
The query was returning wrong data - a lot of jobs had the same reference_workflow_tag_id (which should be unique), the same reference ID but different reference curie values. I guess the old query was missing an ON requirement on the reference id

valearna:
Using more explicit join statement for clarity

sweng66 (approved):
looks good! Thanks @valearna

valearna:
thanks @sweng66 !

PR #817, opened 2025-01-22
Scrum 4344

ianlongden:


sweng66 (dismissed):
Looks good, but can you wait for this one? @ianlongden 
We may need to get rid of A-team API call for getting children of an ATP ID in this sprint.
The example of our own API endpoint looks like:

https://dev4006-literature-rest.alliancegenome.org/topic_entity_tag/search_descendants/ATP:0000006

but in your code, we can call the following function from crud/ateam_db_helpers.py directly:

search_ancestors_or_descendants(ontology_node, ancestors_or_descendants)

search_ancestors_or_descendants("ATP:0000006", "descendants")

sweng66:
Sorry, I accidentally approved code, please do not merge.  @ianlongden

ianlongden:
"_We may need to get rid of A-team API call for getting children of an ATP ID in this sprint_." Yes this makes it crash currently. So will not merge.

PR #818, opened 2025-01-23
updated to replace A-team API calls by direct access A-team persistent store

sweng66:
The changes I made:

1. added the following new API endpoints:

* entity validation
* topic search (for topic typeahead search)
* species search (for species typeahead search)
* search descendants (for searching children for gene, allele, display_tag, …)

3. fixed the API code to replace all a-team API calls with the calls to get id to name / name to id mapping from A-team persistent database (the only thing left is the call to check A-team API is up for alert message - we need to decide if we want to get rid of it).

4. fixed TET UI related code to get rid of A-team API calls - including entity validation, topic and species autocomplete search, taxon to species mapping data - instead, the components/actions are calling ABC API endpoints

5. Sort page is still using A-team API for creating/adding a new species to A-team persistent store so we will leave as is

6. reference merge code: "queryAteamAtpChildren" in mergeActions can be updated to call our own API 'search_descendants' endpoint to get children of an ATP ID, but we can leave this for later.


ianlongden (approved):

sweng66:
Thanks @ianlongden

PR #821, opened 2025-01-24
fixed the entity validation endpoint so it will work when entity name contains either '+' or '/'

sweng66:


valearna (approved):

sweng66:
Thanks @valearna

PR #823, opened 2025-01-28
adding one off script to convert file upload WFT from in progrss (ATP:0000139) to files uploaded (ATP:0000134)

sweng66:
for those final main full text WB files with pdf_type != pdf  (eg, ocr, tif, lib, html, etc)


valearna (approved):

sweng66:
Thanks @valearna

PR #827, opened 2025-01-30
Speed up reports

ianlongden:
The sql changes will speed up the queries a little but this was not the major problem. load_workflow_parent_children was being called with the root_node NOT being passed that meant the default value of the top one was called which meant a full load of the ATPs (with lots of calls) when only the sub parts were needed. SCRUM-4803


valearna (approved):
Thanks for figuring out the ATP issue!

PR #831, opened 2025-02-05
store ATP stuff in globals.

ianlongden:
There are a lot of print statements now in the tests. I think this should be fine as it is only in the tests so will not slow things down and helped me detect what was going on, so I would prefer to keep these in for now.


sweng66 (changes requested):
The "atp_get_children" function in workflow_tag_crud.py is returning the direct children, not all descendants. For example, if we pass "ATP:0000165" to it, it will only return 
['ATP:0000189', 'ATP:0000178', 'ATP:0000166', 'ATP:0000169'], not all ATP IDs under these four trees.

sweng66:
I saw you commented out some logging info. I think they are useful for debugging - they will be showing up in ATP logs. 
Anyway, not important.

ianlongden:
> @ianlongden The "atp_get_children" function in workflow_tag_crud.py is returning the direct children, not all descendants. For example, if we pass "ATP:0000165" to it, it will only return ['ATP:0000189', 'ATP:0000178', 'ATP:0000166', 'ATP:0000169'], not all ATP IDs under these four trees. So maybe we can go over the fixes you made in the code for retrieving data from A-team database together some time this week. Anyway, a lot of change!!! Thanks!!

atp_get_children does just get the direct children. For all descendants use get_all_descendents

ianlongden:
> @ianlongden The "atp_get_children" function in workflow_tag_crud.py is returning the direct children, not all descendants. For example, if we pass "ATP:0000165" to it, it will only return ['ATP:0000189', 'ATP:0000178', 'ATP:0000166', 'ATP:0000169'], not all ATP IDs under these four trees. So maybe we can go over the fixes you made in the code for retrieving data from A-team database together some time this week. Anyway, a lot of change!!! Thanks!!

sweng66:
@ianlongden I see. So can you update the search_crud.py to call "get_all_descendents" instead?

ianlongden:
> @ianlongden I see. So can you update the search_crud.py to call "get_all_descendents" instead?

Done, also cached the results of this as there is some looping etc to get this data and it may be called a lot.

sweng66:
Cool. I will check it out this afternoon. Thanks @ianlongden

sweng66 (dismissed):
Looks good to me! So many changes!! Thanks @ianlongden 
Can you not merge to main yet? @ianlongden 
@valearna @azurebrd can you also take a look? I am not good at the unit tests (so many changes!!).
Also I need to deploy my "hot" fixes to prod after Ceri UIT it tomorrow (for SGD transition)
so it will be nice to merge the code after the 'hot' fixes got into prod.

tests/api/test_reference.py
```diff
@@ -744,97 +764,103 @@ def test_get_bib_info(self, test_reference, auth_headers, test_mod): # noqa
...
+                "species": "NCBITaxon:6239",
+                "topic_entity_tag_source_id": test_topic_entity_tag_source.new_source_id,
+                "negated": False,
+                "novel_topic_data": True,
+                "note": "test note",
+                "created_by": "WBPerson1",
+                "date_created": "2020-01-01"
+            }
+            get_response = client.post(url="/topic_entity_tag/", json=new_tet, headers=auth_headers)
+            print(f"new tet: {get_response.status_code}")
```
azurebrd, line 837: Do we want prints in tests ?  It's fine if we do

sweng66 (approved):

PR #832, opened 2025-02-06
Scrum 4370

azurebrd:


sweng66 (dismissed):
Looks good! Thanks @azurebrd

agr_literature_service/lit_processing/pdf2tei/pdf2tei.py
```diff
@@ -81,13 +90,36 @@ def main():
                     file_upload(db=db, metadata=metadata, file=UploadFile(file=BytesIO(response.content),
                                                                           filename=ref_file_obj.display_name),
                                 upload_if_already_converted=True)
+                    add_to_error_list = False
```
valearna, line 93: If I remember correctly only cases of conversion failure (set to "on_failed") should be reported, but it's not clear from the acceptance criteria. Maybe we should ask Ceri, but it would be nice to add to the report if the error is a grobid error or just a generic one and if the workflow tag is set to "failed" or not.

valearna (approved):
Thanks @azurebrd, the code changes look great. I only have one comment about what types of failures to report.

agr_literature_service/lit_processing/pdf2tei/pdf2tei.py
azurebrd: cool cool, thanks @valearna , will ask at standup

azurebrd: should be fixed now

valearna (approved):
Great, thanks @azurebrd !

azurebrd:
Thanks @valearna !

PR #835, opened 2025-02-12
fixes for SGD transition to use ABC

sweng66:
* turned on SGD generation in PROD ABC
* turned off SGD dqm loading in PROD ABC


azurebrd (approved):
Thanks @sweng66 !

PR #839, opened 2025-02-21
Scrum 4893

azurebrd:


azurebrd:
This is not ready to merge, needs unit tests.  But I'm using
 mod_corpus_association_db_obj: ModCorpusAssociationModel = db.query(ModCorpusAssociationModel).filter_by(mod_id = mod_id, reference_id = reference_id).first()

Instead of .one()  because sometimes there isn't an MCA and I want to check against None, since it's already in a try/except.  I recall that @valearna and @sweng66 were talking about when to use .one() vs .first()  and Shuai suggested using .one() and chatGPT suggested using .first()  so not sure if this isn't good, even though it seems to work.

azurebrd:
[8b5488c](https://github.com/alliance-genome/agr_literature_service/pull/839/commits/8b5488c5854522645ab9d8d28e2f6bed11e634b8)  has working functionality and flake8, but mypy complains

[e5bc5e9](https://github.com/alliance-genome/agr_literature_service/pull/839/commits/e5bc5e9896bdcce4f53abf7d31a41444fd96d338)  fixes mypy but flake8 complains

sweng66:
In mca table, we can also use one_or_none() either one row in the table or none @azurebrd

azurebrd:
Great, that works, thanks @sweng66 !

sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #841, opened 2025-02-24
Update API endpoints to clean up the WFT and TET tags when moving a paper out of corpus and

sweng66:
will return a warning message if manually added TET tags exist. Also removed the mod_corpus_association join code for report APIs.


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #842, opened 2025-02-27
fixed to add the paper to the corpus when setting 'No TET data' statu…

sweng66:
…s AND the paper is not in corpus already


valearna (approved):

azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @valearna @azurebrd

PR #845, opened 2025-03-04
Fixed API/pubmed progran to avoid creating duplicate reference relation rows

sweng66:


valearna (dismissed):
@sweng66 These changes make a lot of sense, thanks! The query to read duplicate relationships is repeated many times. You could probably turn it into a function to make the code more readable and easier to manage, but just a suggestion for refactoring, definitely not needed if you are in a rush with other work

sweng66:
Great suggestion! Will fix the code. @valearna

sweng66:
@valearna Fixed. Can you check again when you are free? Thanks

valearna (approved):
Thanks @sweng66 !

sweng66:
Thanks! @valearna

PR #848, opened 2025-03-05
First draft

ianlongden:


valearna (approved):
Looks good to me @ianlongden. I noticed the new mocked function in the pdf2tei tests. This is because the get_jobs function uses the new methods to retrieve subset relationships?

ianlongden:
> Looks good to me @ianlongden. I noticed the new mocked function in the pdf2tei tests. This is because the get_jobs function uses the new methods to retrieve subset relationships?

Not the get jobs but the get list of classifiers needed to add. Uses a team db so had to be mocked.

PR #849, opened 2025-03-06
Allow passing of mod_abbr to search topic and deal with it acordingly

ianlongden:


sweng66 (approved):
Looks good to me! Thanks @ianlongden

PR #852, opened 2025-03-10
SCRUM-4908

ianlongden:


sweng66 (commented):
Looks good to me. Thanks @ianlongden, but I will defer to @valearna @azurebrd to see if we can move forward.

valearna (approved):

PR #853, opened 2025-03-11
use postgresql extract on date fields to query into bins of time freq…

azurebrd:
…uency by year, month, or week.  date_published_start is not a date, so casting it, not sure what happens if it doesn't cast properly.  output has new field time_period for that binned period.


sweng66:
@azurebrd still reading the code.. but noticed there are a typos for  "EXTRACT(MONTH ...) AS week". For example,

 EXTRACT(MONTH FROM mca.date_updated) AS week

sweng66:
Should we update code (line 636 to line 640) to immediately return an error ("invalid date frequency, use: year, month, week") if an unsupported date_frequency is provided instead of records with an error string?

sweng66:
so date_frequency is only for a single MOD, we don't need to deal with it when mod_abbreviation is not provided. correct?

sweng66:
Maybe we can set the date_column once and build extract dynamically. For example:

date_column = "wt.date_updated"  # default
if date_option == 'reference_created':

-     date_column = "r.date_created"

elif date_option == 'reference_published':

-     date_column = "r.date_published_start::DATE"

elif date_option == 'inside_corpus':

-     date_column = "mca.date_updated"

extract = ""
if date_frequency:

-     if date_frequency == 'year':

-         extract = f", EXTRACT(YEAR FROM {date_column}) AS year"

-     elif date_frequency == 'month':

-         extract = f", EXTRACT(YEAR FROM {date_column}) AS year, EXTRACT(MONTH FROM {date_column}) AS month"

-     elif date_frequency == 'week':

-         extract = f", EXTRACT(YEAR FROM {date_column}) AS year, EXTRACT(WEEK FROM {date_column}) AS week"

sweng66:
@azurebrd You can ignore my comments except the typos. Thanks! A lot of changes!

azurebrd:
Thanks @sweng66 !  Those are all great suggestions, it looks so much better now.  Thanks for catching those typos !  Good point about raising an error and warning whoever's working on it, instead of just returning data and ignoring the bad parameter.  That refactor with parametrization looks so much better, but flake8 still thinks the function is too complex.  It looks pretty readable though.  Thanks for checking the code even though I hadn't requested it yet, you're so on top of things always :)

azurebrd:
Thanks @sweng66 !

PR #855, opened 2025-03-13
fix: relax validation for optional fields in ML model schema to allow…

valearna:
… for null topic

Modified `topic`, `precision`, `recall`, `f1_score`, and similar fields to be optional in the schema and adjusted related logic in the router. These changes ensure more flexibility when handling additional models without topics such as tfidf vectorizers while avoiding validation errors.


sweng66 (approved):
Thanks @valearna

PR #861, opened 2025-03-24
a new script for checking and retrieving the obsolete/deleted entities that are in TET table for all MODs and..

sweng66:
also adding a new endpoint to parse the the QC report and return the content for each mod

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -105,7 +105,7 @@ services:
       ELASTICSEARCH_HOST: "${ELASTICSEARCH_HOST}"
       ELASTICSEARCH_PORT: "${ELASTICSEARCH_PORT}"
       ELASTICSEARCH_INDEX: "${ELASTICSEARCH_INDEX}"
-      LOG_PATH: "${LOG_PATH}"
+      LOG_PATH: "/var/log/automated_scripts/"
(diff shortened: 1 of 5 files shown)


docker-compose.yaml
```diff
@@ -105,7 +105,7 @@ services:
       ELASTICSEARCH_HOST: "${ELASTICSEARCH_HOST}"
       ELASTICSEARCH_PORT: "${ELASTICSEARCH_PORT}"
       ELASTICSEARCH_INDEX: "${ELASTICSEARCH_INDEX}"
-      LOG_PATH: "${LOG_PATH}"
+      LOG_PATH: "/var/log/automated_scripts/"
```
azurebrd, line 108: Should this not stay ${LOG_PATH} for consistency ?

azurebrd (approved):
Thank @sweng66 !  I'm not sure about the docker stuff, it might be good if @valearna also looks at it

docker-compose.yaml
sweng66: purposely do that; otherwise wont work.

sweng66:
Thanks @azurebrd ! @valearna Can you take a look to see if those changes make sense? Thanks

PR #870, opened 2025-04-03
adding more exception catching code

sweng66:


valearna (commented):
Thanks @sweng66 , are these exceptions that should still break the pipeline or it is ok to continue if something happens?

valearna (approved):

sweng66:
Thanks @valearna

PR #873, opened 2025-04-07
hot fix by adding mod in query

ianlongden:
Otherwise it crashed and it was expecting only one result but was getting more if more than one mod had it.


sweng66 (approved):
Looks good! Thanks @ianlongden

PR #882, opened 2025-04-17
fixed to return all topics and associated data for a paper and a mod

sweng66:
plus a new endpoint to return a list newly added PMIDs for a given mod, in the past N days (by default, it is 7 days). THese  new papers got added into MOD corpus, but without a MOD paper ID.


ianlongden (approved):

sweng66:
Thanks @ianlongden

PR #890, opened 2025-04-29
add curation_status table

ianlongden:


sweng66 (approved):
Why only date_created is required? 
date_updated, created_by, & updated_by should also have nullable=False, right?

ianlongden:
> Why only date_created is required? date_updated, created_by, & updated_by should also have nullable=False, right?

This just inherits form the audited model so this is exactly the same as all the others. I presume we always know when its created abut we may not always know by who etc.

PR #895, opened 2025-05-02
refactor(mod_corpus_association): remove add_topic_list function and …

valearna:
…related calls

The function was incorrectly creating curation_status entries with null values, which should be only used for whole paper claiming


azurebrd (approved):
Thanks @valearna !

PR #903, opened 2025-05-14
fix(curation_status): remove unset values from patch object to avoid …

valearna:
…unintentionally set values to None


azurebrd (dismissed):
Thanks @valearna

azurebrd (approved):
Thanks @valearna !

valearna:
thanks @azurebrd !

PR #908, opened 2025-05-21
Scrum 5026 - bulk uploader for FB

valearna:
- Added rule to process FB files
- Names are PMID with numbers only
- No supplemental files for FB, just main pdf files
- Added tests for info extraction rules from file names
- Added script to move uploaded files to specified folder


azurebrd (approved):
Thanks @valearna !

PR #918, opened 2025-06-08
Scrum 5078 new public index

valearna:
@FB-ping Can you check that this PR creates the new public index correctly and make sure it can still run on the aws instances? I've added some github action tests for debezium that I'd like to mark as required for PRs in the future. Please do not merge this PR until we have tested it on the dev server.


sweng66 (dismissed):
I scan over the code and it looks good. Thanks @valearna

FB-ping (approved):

PR #924, opened 2025-06-13
fixed dqm loading script to set the alliance category to 'Other' for FB papers with a 'note' reference type

sweng66:


azurebrd (approved):
Cool, thanks @sweng66 !

PR #925, opened 2025-06-20
New endpoints for bulk upload and status checking endpoints + unit tests

sweng66:
plus a fix to a function in db_write_utils.py to address 'A transaction is already begun on this Session' complaining.


sweng66:
Thanks @ianlongden

PR #928, opened 2025-06-24
Scrum 4850

ianlongden:


sweng66 (changes requested):
priority tag is not workflow, so we can't transition one to another. Can you undo that fix (line 1292)?

ianlongden:
Agree, Removed the PR, problem must be else where.

PR #930, opened 2025-06-24
fixed code to also check other pre-indexing tags before adding needed

sweng66:


sweng66:
@ianlongden Thanks ! =)

sweng66:
Awesome! @ianlongden

PR #932, opened 2025-06-25
SCRUM-4281 New helper script to add atps for a mod and a set of references

ianlongden:
script to copy workflow_transitions with example for ml_model and another,


sweng66 (approved):
Thanks @ianlongden

PR #934, opened 2025-06-25
Scrum 4974

azurebrd:


sweng66 (approved):
Thanks @azurebrd 
I love the detailed comments. Also those cute icons!! =)

PR #935, opened 2025-06-26
Adding a script to check/fix any obsolete PMIDs for each mod and adding a new endpoint to fetch the obsolete PMID data from the report 

sweng66:


sweng66:
@azurebrd Thanks Juancarlos for checking this one. 
Yes, it would be nice to check which PMIDs already checked, but I was too lazy to implement this since the percentage of shared papers is low.

PR #951, opened 2025-07-10
Initial Commit Confidence Levels

pjhale:
Needs cleanup.  Currently only works when a tag is selected.  If not it does nothing.


valearna (approved):
Looks good, thanks @pjhale !

PR #952, opened 2025-07-10
fixed the topic_entity_tag schema file to set the confidence score to two decimal places when adding it into db / displaying it - set this to CREATE/PATCH/SHOW

sweng66:


valearna (approved):

sweng66:
Thanks @valearna

PR #959, opened 2025-07-23
Scrum 5022 - dockerized reverse proxy

valearna:
Added dockerized reverse proxy. Currently running on stage for testing


valearna:
thanks @sweng66 !

PR #961, opened 2025-07-24
replaced the default= and onupdate= column hooks with explicit before_insert/update event listeners

sweng66:


sweng66:
@valearna ChatGPT suggested this fix. Does this fix make sense to you?

valearna (approved):
thanks @sweng66 !

PR #965, opened 2025-08-11
adding a new table 'indexing_priority' for storing ZFIN priority data and adding several api endpoints for retrieving/inserting/updating the priority tags

sweng66:
* also adding an oneoff script for moving data from WFT table to this new table 
* back populating confidence score for each tag


valearna (approved):
thanks @sweng66

PR #967, opened 2025-08-26
First pass at model prod stuff

ianlongden:


ianlongden:
@claude explain how the literature curation system works

PR #968, opened 2025-08-26
Elasticsearch: make authors nested; add Author/ORCID search support

sweng66:
* Convert authors field to nested in ES mapping.
* Update search endpoint to query authors via:
    * analyzed name (authors.name with authorNameAnalyzer)
    * exact name (authors.name.keyword)
    * normalized ORCID (authors.orcid.keyword, supports ORCID:… and plain IDs).
* Fix/adjust authors facet aggregation to work with nested docs (nested agg → inner terms buckets).
* Accept both sort (ES-style) and sort_by_published_date_order (API param).
* Tests updated for nested authors; router adjusted.


azurebrd:
I also don't really know search

PR #973, opened 2025-09-05
feat(revalidation): implement revalidation logic on tag deletion + up…

valearna:
…date validation in case of downstream conflicts


sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #980, opened 2025-09-17
Adding search query normalization and improving  search for longer queries (e.g., full titles) 

sweng66:
The search ranking has been updated to be more robust to punctuation (e.g., :), long pasted titles, and partial typing. It replaces simple_query_string usage with analyzer-based queries and centralizes boost configuration.


valearna:
not sure what this does but it looks good, thanks @sweng66 !

PR #981, opened 2025-09-18
claude first pass at linking ml_model to topic_entity_tag

ianlongden:
Do NOT merge has alembic changes.  will do these just before the commit.


sweng66 (approved):
Thanks @ianlongden

PR #992, opened 2025-09-30
fixed so XREF IDs without a prefix will still work

sweng66:


valearna (approved):
Cool, thanks @sweng66 !

sweng66:
Thanks @valearna

PR #994, opened 2025-10-02
Scrum 5470 - added functions to stop automatic updates of date_updated and updated_by on changes to avoid resetting them when validating tags

valearna:


sweng66 (approved):
Looks great. Thanks @valearna

PR #1124, opened 2026-04-14
fix(auth): remove set_global_user_from_cognito from read endpoints

azurebrd:
Read-only endpoints don't need to look up the user in the DB — they don't write audit fields (created_by/updated_by). Previously, any Cognito-authenticated user whose email had no matching users row would get a 500 ValueError on every endpoint, including reads.

- Remove set_global_user_from_cognito call from ~89 read-only endpoints across 28 router files. Auth is still required via the Security dependency — endpoints remain locked down, but authenticated users no longer need a DB user record to perform reads.
- Change ValueError to HTTPException(403) in user.py so write endpoints return a clear error instead of 500 when no DB user record exists.
- POST endpoints that only read (cross_reference/show_all, referencefile/show_main_pdf_ids_for_curies) also had the call removed.
- GET endpoint that writes (topic_entity_tag/revalidate_all_tags) retains the call since it needs user tracking for audit.
- Add test_user_auth_scenarios.py verifying: authenticated user without DB record can read, gets 403 on write; no auth gets 401 on both.

### DIFF
--- a/agr_literature_service/api/routers/author_router.py
+++ b/agr_literature_service/api/routers/author_router.py
@@ -56,7 +56,6 @@ async def patch(author_id: int,
 def show(author_id: int,
          user: Optional[Dict[str, Any]] = Security(get_authenticated_user),
          db: Session = db_session):
-    set_global_user_from_cognito(db, user)
     return author_crud.show(db, author_id)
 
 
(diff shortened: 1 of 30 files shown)


sweng66 (approved):
Cool. Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1127, opened 2026-04-16
feat(ml_model): add file_classes column to ml_model

valearna:
## Summary
- Add `file_classes` column (string array) to `ml_model` table to specify which file types a model processes
- Alembic migration adds the column and backfills existing rows: `biocuration_topic_classification` → `["main"]`, `biocuration_entity_extraction` → `["main", "supplement"]`
- Fix ML model upload endpoint: accept `file_classes` as comma-separated query string, use `bytes` type for `model_data_file` to handle Swagger UI sending empty strings for optional file fields
- Include standalone backfill script and migration to resolve model-db drift for indexes/foreign keys

## Test plan
- [ ] Verify `file_classes` column is created and backfilled correctly via alembic migration
- [ ] Upload model via Swagger UI with `file_classes=main,supplement` — should store as `["main", "supplement"]`
- [ ] Upload model via JSON metadata file with `file_classes: ["main"]` — should work as before
- [ ] Upload model without selecting `model_data_file` in Swagger — should not return 422
- [ ] Existing ml_model tests pass

### DIFF
--- a/agr_literature_service/api/crud/ml_model_crud.py
+++ b/agr_literature_service/api/crud/ml_model_crud.py
@@ -71,7 +71,8 @@ def upload(db: Session, request: MLModelSchemaPost, file: UploadFile):
         production=request.production,
         species=request.species,
         data_novelty=request.data_novelty,
-        negated=request.negated
+        negated=request.negated,
+        file_classes=request.file_classes
     )
     try:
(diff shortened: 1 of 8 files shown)


sweng66 (approved):
Thanks @valearna

PR #1140, opened 2026-04-29
feat(person): aggregate name search and add cross-reference lookup

azurebrd:
/person/by_name now matches person_name.first/middle/last_name in addition to display_name so curators can find people by structured name parts. Adds /person/by_person_cross_reference/{curie_or_id} mirroring the existing reference/by_cross_reference endpoint. Renames /by/X to /by_X on the person and person_setting routers for consistency with the rest of the API; lookup routes are declared before the {id} catch-all so single-segment paths resolve correctly.

### DIFF
--- a/agr_literature_service/api/crud/person_cross_reference_crud.py
+++ b/agr_literature_service/api/crud/person_cross_reference_crud.py
@@ -46,6 +46,8 @@ def create_for_person(db: Session, person_id: int, payload: Dict[str, Any]) -> P
     if not curie:
         raise HTTPException(status_code=422, detail="curie is required")
 
+    curie_prefix = _curie_prefix(curie)
+
     # curie is globally unique on person_cross_reference (uq_person_xref_curie).
     dup = (
         db.query(PersonCrossReferenceModel.person_cross_reference_id)
@@ -58,10 +60,28 @@ def create_for_person(db: Session, person_id: int, payload: Dict[str, Any]) -> P
             detail=f"Cross-reference '{curie}' already exists",
         )
 
+    # (person_id, curie_prefix) is unique per-person (uq_person_xref_person_prefix).
+    prefix_dup = (
+        db.query(PersonCrossReferenceModel.person_cross_reference_id)
+        .filter(
+            PersonCrossReferenceModel.person_id == person_id,
+            PersonCrossReferenceModel.curie_prefix == curie_prefix,
+        )
+        .first()
+    )
+    if prefix_dup:
+        raise HTTPException(
+            status_code=422,
+            detail=(
+                f"Another cross-reference with prefix '{curie_prefix}' "
+                f"already exists for this person."
+            ),
+        )
+
     obj = PersonCrossReferenceModel(
         person_id=person_id,
         curie=curie,
-        curie_prefix=_curie_prefix(curie),
+        curie_prefix=curie_prefix,
         pages=_clean_pages(data.get("pages")),
         is_obsolete=bool(data.get("is_obsolete", False)),
     )
--- a/agr_literature_service/api/crud/person_cross_reference_crud.py
+++ b/agr_literature_service/api/crud/person_cross_reference_crud.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, List
 from fastapi import HTTPException, status
 from fastapi.encoders import jsonable_encoder
 from sqlalchemy import or_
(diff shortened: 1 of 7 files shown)


azurebrd:
Re: "Bug (minor, harmless in practice)"  : "The current code intentionally mirrors cross_reference_crud.get_cross_reference (crud/cross_reference_crud.py:28-37) and person_crud.resolve_person_id (crud/person_crud.py:30-47). All three use the same or_(curie == X, id == None_or_int) pattern.  Fixing only get_by_curie_or_id introduces a one-off inconsistency with its two siblings. "

azurebrd:
Re: "find_by_name now runs ILIKE '%pattern%' across person_name.first_name, middle_name, and last_name. There are no indexes on those columns, and leading-wildcard ILIKE can't use btree indexes anyway, so this will seq-scan the join. This matches the pre-existing pattern on display_name and is fine at current scale. If the person_name table grows large and these lookups become frequent, a pg_trgm GIN index would help."

" the existing index named
  ix_person_display_name_trigram (alembic/versions/20251008_a3e751dff3e2...py:206-208) is misleadingly named — it's a plain btree index (no
  postgresql_using="gin", no gin_trgm_ops), so it can't help with leading-wildcard ILIKE either. The pre-existing display_name search has been
  seq-scanning all along."
  
  For less than 200k to scan, it's < 250ms   
  At 2M to 10M   1-5s
  Recommends leaving as-is until there's a lot more entries, or if we feel like doing it now, an alembic file with
  
    def upgrade():
      op.execute("CREATE EXTENSION IF NOT EXISTS pg_trgm")

      # Replace the misnamed btree with a real GIN trigram index
      op.drop_index("ix_person_display_name_trigram", table_name="person")
      op.create_index(
          "ix_person_display_name_trgm", "person", ["display_name"],
          postgresql_using="gin",
          postgresql_ops={"display_name": "gin_trgm_ops"},
      )

      # Add GIN trigram indexes to person_name fields
      for col in ("first_name", "middle_name", "last_name"):
          op.create_index(
              f"ix_person_name_{col}_trgm", "person_name", [col],
              postgresql_using="gin",
              postgresql_ops={col: "gin_trgm_ops"},
          )

  def downgrade():
      # symmetric drops, restore the old btree
      ...

Edit:  Actually we need that for person_name also : 
if you go the trigram route, you need it on person_name's columns too, not just display_name.

  The reason is how Postgres uses indexes with OR. The new find_by_name query is:

  WHERE display_name ILIKE '%X%'
     OR first_name ILIKE '%X%'
     OR middle_name ILIKE '%X%'
     OR last_name ILIKE '%X%'

  Postgres can combine multiple GIN scans via a BitmapOr — but only if each branch has its own usable index. If you put a GIN trigram on display_name
  only and leave the three person_name columns unindexed, the planner can't use BitmapOr; it falls back to a full scan of the joined table because the
   un-indexed branches force it. Net effect: the new aggregation query is no faster than today.

  So to actually accelerate the new find_by_name, you'd index all four columns:
  - person.display_name (replacing the misnamed btree)
  - person_name.first_name
  - person_name.middle_name
  - person_name.last_name

  Cost is small. GIN trigram indexes don't store NULL trigrams, so a mostly-NULL column like middle_name adds almost no disk. At 200K total
  person_name rows, the four indexes combined would land somewhere under 10 MB. Write overhead per insert/update on these columns: 3 GIN entries per
  indexed column (one per trigram in the value) — fine for a low-churn table.

  Bonus: person_setting_crud.find_by_name now uses the exact same OR-of-four-columns shape, so the same indexes serve both queries without extra work.

  So the answer is: it's all-or-nothing. Either you add real GIN trigram indexes on all four columns and both find_by_name paths get fast, or you
  don't and they keep seq-scanning. Adding it only on display_name is an in-between state that buys you nothing for the new aggregation.

valearna:
@azurebrd this PR looks great, thanks. I recommend changing the index type for all name fields to GIN (generalized inverted index), which will make search faster (write will be slower, but we are probably doing more reads than writes to the person table). We are already using indexes for all these fields but it's the default BTREE type, not optimized for ILIKE searches

azurebrd:
@claude review again

azurebrd:
@claude review again

sweng66 (approved):
Thanks for these changes. @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1146, opened 2026-05-05
autocomplete_on_id function had inefficient ilike search and CC wante…

azurebrd:
…d to change the day it's indexed, but looking at how it's used, it's not used other than by the router, and the UI does not call it, and there are zero uses of it in the nginx logs on prod and stage, so removing the endpoint and function


azurebrd:
@claude review changes since last review

azurebrd:
@claude review again

azurebrd:
@claude review again

sweng66 (approved):
Wow! so many changes! Thanks @azurebrd

PR #1147, opened 2026-05-06
SCRUM-6014 limit supplemental PDF conversion size

sweng66:


sweng66:
@claude can you check code?

sweng66:
From codex:

  The conservative unknown-page-count behavior is intentional for this change: if we cannot
  confidently prove a supplemental PDF is within the 20-page cap, we do not send it to PDFX.
  The /Count fallback is deliberately second-choice after pypdf and /Type /Page scanning.

  One note: line length should be okay because this repo’s setup.cfg has E501 ignored, but
  make run-local-flake8 may still be worth running in a fully provisioned env before merge.

valearna (approved):
looks good, thanks @sweng66 !

PR #1152, opened 2026-05-11
preserve emails with reference relations on person/email deletion and add retraction_status to search endpoint

sweng66:
- Change email.person_id FK from CASCADE to SET NULL
- Remove cascade delete-orphan from PersonModel.emails relationship
- Add check in email_crud.destroy() to prevent deletion of emails with reference relations
- Update person_crud.destroy() to detach emails with reference relations instead of deleting
- Add migration to update FK constraint in database

This ensures that emails linked to references are preserved when:
- A person is deleted (email becomes orphan with person_id=NULL)
- An email is directly deleted (returns 422 error instead)

### DIFF
--- a/agr_literature_service/api/crud/email_crud.py
+++ b/agr_literature_service/api/crud/email_crud.py
@@ -8,7 +8,7 @@ from fastapi import HTTPException, status
 from fastapi.encoders import jsonable_encoder
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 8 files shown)


sweng66:
@claude Can you check code again?

sweng66:
@claude can you take a look?

azurebrd (approved):

sweng66:
Thanks @azurebrd

PR #1153, opened 2026-05-13
fix(api): tighten schema validation to reject unknown PATCH/POST fields

azurebrd:
SCRUM-1899: PATCH/POST bodies with bogus field names silently returned 202/201 instead of erroring. Switch the affected schemas to Pydantic's extra='forbid' so unknown fields produce a 422.

- CopyrightLicenseSchemaPost: extra='ignore' -> 'forbid'. Was originally forbid; loosened in acb664cd "fixing unit test errors" in 2025-07. Current tests only send the four real fields, so the original failure no longer reproduces.
- AuditedObjectModelSchema (base): extra='ignore' -> 'forbid' as a safer default for future schemas. Every existing subclass declares its own model_config, so behavior is unchanged today.

FacetsOptionsSchema (search filters) intentionally left as 'ignore' -- it absorbs legacy field names by design.

### DIFF
--- a/agr_literature_service/api/schemas/base_schemas.py
+++ b/agr_literature_service/api/schemas/base_schemas.py
@@ -4,7 +4,7 @@ from pydantic import BaseModel, ConfigDict
 
 
 class AuditedObjectModelSchema(BaseModel):
-    model_config = ConfigDict(extra="ignore", from_attributes=True)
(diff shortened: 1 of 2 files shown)


sweng66 (approved):
Thanks @azurebrd

PR #1154, opened 2026-05-13
Add literature reference fuzzy search indexes

christabone:
## Summary
- add expression GIN `pg_trgm` indexes for literature reference validation lookups
- cover partial/case-insensitive matches on `reference.curie`, `reference.title`, and active `cross_reference.curie`
- ensure `pg_trgm` exists before creating the validation indexes

## Validation
- `git diff --check`
- `python3 -m py_compile alembic/versions/20260513_b1c2d3e4f5a6_add_reference_validation_trigram_indexes.py`
- pre-commit secret scan: Gitleaks passed
- pre-commit secret scan: TruffleHog passed

## Context
Live read-only literature DB inspection showed btree indexes on `reference.curie` and `cross_reference.curie`, but no trigram indexes for the `UPPER(...) LIKE` predicates used by the AI curation validation helper searches.

### DIFF
--- a/alembic/versions/20260513_b1c2d3e4f5a6_add_reference_validation_trigram_indexes.py
+++ b/alembic/versions/20260513_b1c2d3e4f5a6_add_reference_validation_trigram_indexes.py
@@ -1,0 +1,50 @@
+"""add reference validation trigram indexes
+
+Revision ID: b1c2d3e4f5a6
+Revises: 0a43280cf638
+Create Date: 2026-05-13
+
+Add expression GIN trigram indexes for AI curation validation lookups that
+match literature references by partial titles and paper-derived identifiers.
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "b1c2d3e4f5a6"
+down_revision = "0a43280cf638"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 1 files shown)


valearna:
@christabone looks good to me. We have added similar indexes for other fields. @azurebrd @sweng66 can we apply these alembic changes to stage and merge? This is similar to the changes we have added to the person table. Chris needs a quick way to search for references and pull out ref info from our db with searches on title and curie using the unified library (the same we are using to connect to the persistent store). We might move this into a dedicated api in our backend at some point, which could use elasticsearch for the indexes but I don't think it hurts to have the GINs on the ref columns.

valearna:
@christabone  I'm closing this one — we ended up taking a different route for the curation API client's literature reference search.

In alliance-genome/agr_curation_api_client#17 the `search_literature_references` / `get_literature_reference` helpers now query the literature `references_index` on OpenSearch directly instead of running `UPPER(...) LIKE '%q%'` against Postgres. The fuzzy/partial behavior the trigram GIN indexes were meant to support is now handled by ES (`fuzziness: AUTO` on `title`/`citation`/`abstract`, `wildcard` on `curie.keyword` and `cross_references.curie.keyword`), so the ABC DB doesn't need the new indexes.

Thanks for putting this together — happy to revisit if another caller turns out to need trigram support on those columns.

PR #1162, opened 2026-05-17
feat(SCRUM-5867): replace pdf2tei cron with pdf2md

valearna:
## Summary
- Swaps the weekly Thursday 02:00 \`crontab\` entry from \`pdf2tei.py\` to \`pdf2md.py\` — \`pdf2md.main()\` already iterates the \`text_convert_job\` workflow queue and emits the same \`job_change_atp_code\` on_success/on_failed transitions (ATP:0000163) that \`pdf2tei\` used to drive, so the workflow handoff carries over unchanged.
- Drops the transient \`is_converted_output\` short-circuit added to \`referencefile_crud.file_upload()\` under SCRUM-5795. With every downstream pipeline now consuming Markdown, entity/classification jobs no longer overlap with the conversion run, so \`converted_*\` uploads can flow through the same \`is_file_upload_blocked\` guardrail as any other file_class. \`pdf2md\` itself uploads while the workflow tag is still at "text conversion needed" (ATP:0000162) — not in_progress — so the guardrail does not block its own writes.
- Removes \`tests/lit_processing/pdf2tei/\`. The \`pdf2tei.py\` module itself is retained in the tree for reference.

## Test plan
- [ ] Confirm \`pdf2md.py\` runs end-to-end against staging and produces \`job_change_atp_code\` transitions matching the prior \`pdf2tei\` behavior
- [ ] Verify a \`converted_*\` upload while a reference has no conflicting in-progress workflow still succeeds (not blocked by the restored \`is_file_upload_blocked\` guardrail)
- [ ] Verify a \`converted_*\` upload IS now correctly blocked when an overlapping workflow is in progress (regression check for the removed short-circuit)
- [ ] After merge: rebuild the automated-scripts container and \`make restart-automated-scripts\`; confirm the new \`/var/log/automated_scripts/pdf2md.log\` is being written and that \`pdf2tei.log\` no longer rotates


azurebrd (approved):

valearna:
thanks @azurebrd !

PR #1164, opened 2026-05-18
fix(SCRUM-6092): transition text_convert_job tag when supplements fail

valearna:
## Summary

- **Bug:** The on-demand conversion path left the per-MOD `text_convert_job` workflow tag pinned to "needed" whenever any supplement remained unconverted — including supplements that were permanently rejected (>5 MB / page-count limit) or transiently failed (GROBID empty, PDFX errors). Real fallout: a Friday upload with a >5 MB supplemental left the workflow stuck even though the main converted cleanly.
- **Cause:** `transition_completed_text_convert_tags` keyed the transition off `all_converted`, which required `pending_supplement_count == 0`. A skipped supplement leaves no `converted_merged_supplement` row and is indistinguishable from one that was never attempted.
- **Fix:** New `mod_has_converted_main` helper (affirmative DB check for a `converted_merged_main` row matching one of the MOD's main sources), surfaced via `per_mod_pending_status` as `main_converted`. The tag transition now keys off `main_converted` — supplement failures no longer block the workflow. This brings the on-demand path in line with the cron path (`process_single_reference` already returns success on main-only).
- **API:** `ConversionModStatusSchema` gains a `main_converted: bool` field. The existing `all_converted` is preserved (slightly narrowed: now means "main converted AND supplements converted" instead of "no pending of either kind"); callers that want the full picture can still read it.

## Files changed

| File | Change |
|---|---|
| `lit_processing/pdf2md/pdf2md_utils.py` | New `mod_has_converted_main(...)` helper. |
| `api/crud/file_conversion_crud.py` | `per_mod_pending_status` now returns `main_converted`. `transition_completed_text_convert_tags` keys off `main_converted` instead of `all_converted`. |
| `api/schemas/file_conversion_schemas.py` | `main_converted: bool` added to `ConversionModStatusSchema`. |
| `api/utils/conversion_processor.py` | Comment updated to reflect main-only success semantics. |
| `tests/api/test_file_conversion.py` | New `TestSupplementFailureTagTransition` class (3 tests). |

## Test plan

- [x] `make run-local-flake8` — clean
- [x] `make run-local-mypy` — clean (521 source files)
- [ ] CI: `tests/api/test_file_conversion.py::TestSupplementFailureTagTransition` (three new tests covering: schema exposes `main_converted`; tag lands when main converted + supplement pending; tag held when main still pending)
- [ ] CI: existing `tests/lit_processing/pdf2md/test_pdf2md.py::TestProcessSingleReference::test_supplement_failures_do_not_fail_reference` still passes (cron-side behavior unchanged)
- [ ] Manual on devserver: upload a reference with a converting main + a >5 MB supplemental, call the on-demand endpoint, confirm the MOD's `text_convert_job` tag transitions to "conversion done" while the supplement is reported as failed in `per_file_progress`

## Related

- KANBAN-1327 — upstream pdfx-side complement (don't abort the whole pipeline when a single extraction method fails). This PR is the ABC-side fix that lets the workflow tag move regardless of the upstream's per-method robustness.

### DIFF
--- a/agr_literature_service/api/crud/file_conversion_crud.py
+++ b/agr_literature_service/api/crud/file_conversion_crud.py
@@ -32,6 +32,7 @@ from agr_literature_service.lit_processing.pdf2md.pdf2md_utils import (
     get_nxml_referencefile,
     get_pdf_files_for_reference,
     is_eligible_for_supplement_conversion,
+    mod_has_converted_main,
     pending_main_sources,
     pending_supplement_sources,
     process_nxml_to_markdown,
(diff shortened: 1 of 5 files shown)


sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #1169, opened 2026-05-20
fix(SCRUM-6106): isolate _current_user_id per request via ContextVar

valearna:
## Summary

Fixes [SCRUM-6106](https://agr-jira.atlassian.net/browse/SCRUM-6106).

- Replace the module-level `_current_user_id` global in
  `agr_literature_service/api/user.py` with a
  `contextvars.ContextVar`, isolating the current user per request
  across both asyncio tasks and FastAPI's sync-endpoint threadpool.
- The SQLAlchemy `before_update` listener on `AuditedModel` reads
  this value to stamp `updated_by`; under the old global, concurrent
  curator requests could trample each other and produce wrong
  attributions (the symptom that prompted the ticket was SGD
  `mod_corpus_association` rows showing an FB curator as the
  "sorter").
- Public API (`get_global_user_id` / `set_global_user_id` /
  `set_global_user_from_cognito` / `get_current_user_pk`) is
  unchanged, so no router or event-listener call site needs to
  change. Function names retain the "global" prefix for now to keep
  the diff small; a rename across the ~244 call sites is a follow-up.

## Tests

- `tests/api/test_user_context_isolation.py` (new):
    - asyncio-task race: two `asyncio.gather` tasks set different
      ids with overlapping sleeps; each must observe its own.
    - threading race: two raw `threading.Thread` workers,
      coordinated by a `Barrier`, must observe their own ids — a
      guard against re-introducing a process-global.
    - threadpool reuse: two sequential `anyio.to_thread.run_sync`
      calls; the second must observe `None` (the first call's
      `set()` is discarded by anyio's `copy_context().run(...)`).
      This is the actual dispatch path FastAPI uses for sync
      endpoints.
- `tests/api/test_audited_model.py`: existing `_clear_global_user`
  autouse fixture updated to use the ContextVar set/reset token
  API (previously assigned to the module attribute directly).

All three tests fail under the pre-fix global and pass under the
ContextVar.

## Historical data

Out of scope for this PR. The `mod_corpus_association_version`
table stores the same corrupted `updated_by` value, so the original
correct curator cannot be recovered from the database. Curators
will review and correct suspect rows manually.

## Test plan

- [x] `make run-local-flake8` — clean.
- [x] `make run-local-mypy` — `Success: no issues found in 527 source files`.
- [x] Manual smoke test on RDS dev: sorting a paper now stamps the
      acting curator on `mod_corpus_association.updated_by`
      (verified on `AGRKB:101000001224742`).
- [ ] CI: full test suite, including the new
      `test_user_context_isolation.py` regression cases.
- [ ] Manual concurrency check post-merge: two curators (or one
      curator + an automated job) operating on the dev API in
      parallel — confirm `updated_by` always matches the
      requesting curator.

[SCRUM-6106]: https://agr-jira.atlassian.net/browse/SCRUM-6106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agr_literature_service/api/user.py
+++ b/agr_literature_service/api/user.py
@@ -42,8 +42,8 @@ def _ensure_automation_user(db: Session, program_name: str) -> UserModel:
 
 def set_global_user_id(db: Session, id: str) -> None:
     """
-    Manually set the global users.id (e.g., script/program name).
-    Treated as an automation user.
+    Set the current users.id for this request/context (e.g., script or
+    program name). Treated as an automation user.
     """
     _current_user_id.set(id)
     _ensure_automation_user(db, id)
@@ -58,7 +58,7 @@ def add_user_if_not_exists(db: Session, user_id: str) -> None:
 
 def set_global_user_from_cognito(db: Session, cognito_user: Optional[Dict[str, Any]]) -> None:
     """
-    Set the global user from a Cognito token.
+    Set the current request's user from a Cognito token.
 
     For ID tokens (user login): Looks up user by email via email table join.
     For access tokens (service accounts): Uses 'default_user' and creates if needed.
--- a/agr_literature_service/api/user.py
+++ b/agr_literature_service/api/user.py
@@ -1,3 +1,4 @@
+from contextvars import ContextVar
 from typing import Optional, Dict, Any
 from fastapi import HTTPException
 from sqlalchemy import text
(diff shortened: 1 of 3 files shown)


sweng66 (approved):
Thanks for the quick fix @valearna

PR #1172, opened 2026-05-21
always add human papers to AGR mod regardless of other MOD corpus

sweng66:
Previously, human papers from GOA, GAF, and interaction datasets were only added to AGR if they weren't in any other MOD's corpus. This change ensures human papers are always added to the AGR mod if not already present there, even if they exist in other MOD corpora.

- Add `add_even_if_in_other_corpus` parameter to `associate_papers_with_alliance()`
- Update load_goa_human_papers.py to use the new parameter
- Update load_mod_gaf_papers.py (process_human_gaf) to use the new parameter
- Update load_interaction_papers.py for HUMAN dataset to use the new parameter


azurebrd (approved):
Thanks @sweng66

PR #1180, opened 2026-05-28
feat(SCRUM-5716): Phase 1 Group 1 — strict-REST migration (quick wins)

azurebrd:
Phase 1 Group 1 of the SCRUM-5716 API-output standardization. Migrates the low-risk endpoints to the new strict-REST convention: POST returns the full created object + Location header, PATCH returns the full updated object at 200, DELETE returns 204 with empty body.

Backend changes:
- POST /author/: response_model=AuthorSchemaShow, returns full Author object, sets Location header. author_crud.create now returns the ORM object instead of the bare id.
- POST /curation_status/: response_model=CurationStatusSchemaShow, full object + Location. PATCH /curation_status/{id}: 202 -> 200, response_model=CurationStatusSchemaShow, returns full object. curation_status_crud.create/patch return the ORM object.
- PATCH /workflow_tag/{id}, /manual_indexing_tag/{id}, /reference_relation/{id}, /indexing_priority/{id}: bare int -> 200
  + full ...SchemaShow. Each route now calls the entity's existing CRUD .show() helper after PATCH to return the enriched response (with reference_curie / mod_abbreviation / updated_by_email populated, matching the GET shape).
- DELETE /reference/referencefile/{id}: 200 + 'success' string ->
  204 with empty body, matching every other DELETE.

New utilities:
- agr_literature_service/api/util/resource_urls.py with author_url() and curation_status_url() helpers for Location headers. Pattern will extend as more CREATE endpoints migrate.
- New CurationStatusSchemaShow in agr_literature_service/api/schemas/curation_status_schemas.py exposing curation_status_id + the fields the existing show() already returns.

Test updates (tests/api/):
- test_author.py, test_curation_status.py: extract the new id from the response object (response.json()['author_id'] / response.json()['curation_status_id']) instead of expecting a bare scalar return.
- test_curation_status.py PATCH: expect 200 instead of 202.
- test_workflow_tag.py, test_manual_indexing_tag.py, test_ref_relation.py PATCHes: expect 200 instead of 202.
- test_indexing_priority.py untouched (CRUD-level tests only).
- test_referencefile.py untouched (no direct DELETE-via-HTTP test for the migrated route).

Lint: flake8 + mypy clean on touched files (no new issues introduced beyond the existing pre-existing baseline). UI is already prepared to consume the new shapes via the Phase 0 shim (commit 740ba4f in lit_biblio_ui).


azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

sweng66 (approved):
wow! so many files updated. Thanks @azurebrd 
 
* POST endpoints now return full objects instead of just IDs
* PATCH endpoints changed from 202 Accepted to 200 OK and also return full object?
* DELETE endpoints changed to 204 No Content

That is cool! 

Since you have also updated normalize_email() to catch ValueError in this PR, do you think we should also add try/except blocks for normalize_email in the following 3 CRUD functions?

  - person_email_crud.py:create_for_person (line 50)                                                                        
  - person_email_crud.py:patch (line 146)                                                                                   
  - person_crud.py:create (line 141)    

It looks great! so feel free to ignore this suggestion if I missed anything.

azurebrd:
@sweng66 CC says
  The reviewer's intuition is right that normalize_email can raise ValueError, but at those three call sites the value coming in has already been
  validated by Pydantic at the schema layer:

  - person_email_crud.create_for_person:50 — receives data from a PersonEmailSchemaCreate parse, which runs @field_validator("email_address") calling
  the same _normalize_email in schemas/person_email_schemas.py:23-26. By line 50 it's already either rejected as 422 or a valid string.
  - person_email_crud.patch:146 — same story for PersonEmailSchemaUpdate (schemas/person_email_schemas.py:40-43).
  - person_crud.create:141 — child emails come in as PersonEmailSchemaCreate items (via the parent PersonSchemaCreate), so the same field validator
  ran.

  The reason reference_crud.set_reference_emails:536 and add_reference_email:583 do wrap in try/except is that those code paths accept raw strings
  (from non-validating callers); the schema-level validator isn't in their chain.

  Adding try/except in the three places the reviewer flagged would be belt-and-suspenders against a scenario Pydantic already guarantees can't reach
  the CRUD. Per the project's "don't validate at internal boundaries" guideline, I'd leave them alone. If you ever change a route to bypass
  PersonEmailSchemaCreate/...Update, then the CRUD becomes a real boundary and the try/except is warranted — but that's a future-state decision tied
  to the routing change, not this PR.

  Counterpoint worth considering: consistency. If you want every normalize_email call site to look identical, add the try/except to all three. The
  pattern would then be uniform across reference_crud and person_*_crud, at the cost of three small dead-code branches. Either stance is defensible;
  my recommendation is leave as-is, but I won't push back if you'd rather add them for uniformity.

valearna (approved):
Looks great, thanks @azurebrd !

PR #1185, opened 2026-06-02
prevent duplicate email-extraction workflow tags

sweng66:
proceed_on_value blindly inserted an initial workflow tag (e.g. "email extraction needed") every time its action fired. Re-running a transition (such as a backfill re-setting "file converted to text") therefore added a duplicate "needed" tag alongside an existing "complete" tag, leaving conflicting states the workflow code cannot reconcile.

Add an idempotency guard that snapshots which workflow processes the reference+mod already has (keyed by process/parent ATP) BEFORE inserting, and skips seeding any already-present process. The pre-insert snapshot preserves legitimate multi-tag seeding within a single action call (a process "needed" tag plus its subtask "needed" tags).

Also add a one-off repair script that dynamically finds and fixes papers with duplicate or missing email-extraction tags. The missing-tag detection is scoped to recently corpus-added papers (--since-date, default 2025-01-01) to match the email-extraction rollout window; it is dry-run by default and requires --commit to apply.

### DIFF
--- a/agr_literature_service/api/crud/workflow_transition_actions/proceed_on_value.py
+++ b/agr_literature_service/api/crud/workflow_transition_actions/proceed_on_value.py
@@ -1,6 +1,10 @@
 from agr_literature_service.api.models import WorkflowTagModel
 from fastapi import HTTPException, status
-from agr_literature_service.api.crud.ateam_db_helpers import get_workflow_tags_for_mod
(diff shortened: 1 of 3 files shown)


sweng66:
@claude Can you check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #1187, opened 2026-06-04
nclude indexing_priority tables in Debezium publication

sweng66:
The unified connector uses publication.autocreate.mode=filtered, which creates the publication only when absent and never adds newly-included tables to a pre-existing one. So indexing_priority and manual_indexing_tag (added to table.include.list later) were never published, and their row changes — including curator_indexing_priority — never streamed into the search index.

- setup.sh: drop publications alongside replication slots so the connector rebuilds them against the current table.include.list on every reindex.
- heal_publication.sh: idempotent, non-destructive self-heal that ALTERs the existing publication to add any missing tables, so a running deployment starts streaming them without a full reindex.
- debezium_setup.dockerfile: ship heal_publication.sh in the dbz_setup image.
- Makefile: add heal-debezium-publication target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,13 @@ restart-debezium-aws:
 stop-debezium:
 	docker-compose --env-file ${ENV_FILE} rm -svf dbz_connector dbz_kafka dbz_zookeeper dbz_ksql_server dbz_setup
 
+# Non-destructive self-heal: add any tables in table.include.list that are missing from the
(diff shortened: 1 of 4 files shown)


valearna (approved):
Thanks @sweng66, dropping the publication seems the best solution. I would restart debezium to make sure the restart script picks up the updated publication with the additional tables

sweng66:
Many thanks @valearna

PR #1191, opened 2026-06-08
protect curator-added reference_email rows from automation

sweng66:
set_reference_emails now keys its replace behavior off the acting user (get_global_user_id). A real human curator (id starts with 'AGRKB:103', person-backed) keeps full control — add/update/delete via a clean full replace. A non-human actor (the email-extraction pipeline / service accounts running as default_user) can no longer remove or overwrite curator-added rows: rows whose updated_by starts with 'AGRKB:103' are preserved, and an incoming address duplicating one is skipped.


valearna (approved):
thanks @sweng66 !

sweng66:
Thanks @valearna  can you re-approve?

PR #1194, opened 2026-06-08
add species to obsolete entity report

sweng66:
Surface the species of each deleted/obsoleted TET entity so curators can tell whether to expect the entity in the A-team database.

- report_obsolete_entities.py: aggregate distinct tet.species per entity, resolve NCBITaxon curies to names via the A-team DB, and append species as a trailing column in the report log
- check_crud.py: parse the new species field into the /check/check_obsolete_entities response (empty for older log files)


valearna (approved):

sweng66:
Thanks @valearna

PR #1195, opened 2026-06-10
support source method and SEA exclusion in TET search

sweng66:
Generalize add_tet_facets_values to emit top-level must_not nested clauses for source_method / source_evidence_assertion exclusions (whole-reference semantics: drop a reference if any tag matches). Confidence level negation behavior is unchanged. Guard the empty-filter cleanup so a standalone exclusion is not dropped, and run the helper when only negative values exist.


sweng66:
@claude check code again?

azurebrd (approved):

sweng66:
@azurebrd Thanks!

PR #1196, opened 2026-06-12
Scrum 6089

azurebrd:


azurebrd:
@claude review again

azurebrd:
@claude review

sweng66 (approved):
Wow, 60 files changed. I’m still trying to understand the purpose of allele_designation. Is this mainly WB-specific?

Also, should an allele_designation be unique per MOD across labs? From what I can tell, the current constraint enforces only one designation per (lab, MOD), but it does not make the designation/code itself unique. Just wanted to confirm whether that is intentional. @azurebrd Anyway, just a question.

PR #1197, opened 2026-06-15
auto-create users from created_by/updated_by on all writes

sweng66:
Auto-create the users referenced by created_by/updated_by at the AuditedModel before_insert/before_update listener level, so an explicit "created by"/"updated by" name supplied to any POST/PATCH/write endpoint (e.g. an admin-token tag load) no longer violates the users.id foreign key. Previously this auto-create was wired only into the topic-entity-tag CRUD.

Add ensure_user_exists_on_connection() in api/user.py: a Connection-based counterpart to add_user_if_not_exists that emits INSERT ... ON CONFLICT DO NOTHING on the in-flight flush connection, satisfying the FK within the same transaction without an extra commit.

### DIFF
--- a/agr_literature_service/api/models/audited_model.py
+++ b/agr_literature_service/api/models/audited_model.py
@@ -4,7 +4,7 @@ import pytz
 from sqlalchemy import Column, ForeignKey, DateTime, event
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.inspection import inspect as sa_inspect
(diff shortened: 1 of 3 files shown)


sweng66:
@claude Can you check again?

sweng66:
@claude check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #1198, opened 2026-06-16
fix(SCRUM-6089): pin safe Swagger examples for laboratory PATCH email…

azurebrd:
…_visibility/lab_is_open

LaboratorySchemaUpdate fields are optional (omit = leave unchanged), but with no example pinned Swagger filled the generated PATCH body with the first Literal value ("public") and true for the bool — implying a partial update would set the lab public and open. Pin examples to the safe POST defaults (not_shown / false) so the prefilled "Try it out" body is no longer misleading. CV and create defaults are unchanged.


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1201, opened 2026-06-17
auto-create companion entity tag for positive mixed tags AND match entity_type by ATP hierarchy in validation

sweng66:
SCRUM-6183 — Auto-create companion entity tag for mixed tags

When a curator creates a positive mixed topic+entity tag via the API, also create a companion "pure entity" tag (topic == entity_type) with data_novelty "existing data" (ATP:0000334), so curators don't have to add it by hand.

- create_entity_tag_for_mixed_tag(): idempotent, non-fatal helper that reuses the duplicate-check path and never rolls back the committed parent tag.
- Gated behind validate_on_insert so it only fires on the interactive create path, not bulk reference import / reference-merge copies. SGD is excluded.
- Back-population one-off script for historical mixed tags (excludes SGD).
- Tests for companion creation, negated/pure-entity/SGD exclusions.

SCRUM-6188 — Match entity_type by ATP hierarchy in validation
  
  TET validation traversed the ATP ontology for topic and data_novelty but compared entity_type with exact string
  equality, so a more specific manually-entered tag (e.g. classical allele ATP:0000285) failed to validate the
  generic allele tag (ATP:0000006) for the same entity.
  - Added atp_hierarchy_with_self() and made every cross-tag entity_type comparison in all three validation helpers
  (and their entity-only fallback branches) hierarchy-aware, following the same generic/specific direction already
  used for topic/data_novelty.
  - Entity string still matches exactly; same-tag pure-entity/mixed discriminators stay exact; exact entity_type
  matches still validate.
  - Back-population: re-run revalidate_all_tags after deploy to link existing references (e.g.
  AGRKB:101000000900959).
  
  Tests
  
  New tests for companion creation (positive/negated/pure-entity/SGD cases) and hierarchy-aware entity_type
  validation; updated one existing count assertion (6→7).

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -171,12 +171,13 @@ def create_tag(db: Session, topic_entity_tag: TopicEntityTagSchemaPost,
             # reference-merge copies, which pass validate_on_insert=False. SGD is excluded
             # (it has its own data_novelty/display handling).
             is_mixed = (topic_entity_tag_data.get("entity") is not None
+                        and topic_entity_tag_data.get("entity_type") is not None
                         and topic_entity_tag_data.get("entity_type") != topic_entity_tag_data["topic"])
             is_positive = topic_entity_tag_data.get("negated") is False
             is_sgd = source.secondary_data_provider.abbreviation == "SGD"
             if is_mixed and is_positive and not is_sgd:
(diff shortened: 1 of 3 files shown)


sweng66:
@claude Can you check again?

sweng66:
@claude Can you check again?

sweng66:
@claude Can you check again?

sweng66:
@claude Final check?

azurebrd (approved):

sweng66:
Thanks @azurebrd

PR #1202, opened 2026-06-17
feat(SCRUM-5819): laboratory name/strain lookup + lab_person display …

azurebrd:
…fields

- Fix the mock laboratory curie prefix AGRKB:705 -> AGRKB:704 (CURIE_PREFIX, doc comments, test expectation).
- Add GET /laboratory/by_name_or_strain_designation?query= backed by laboratory_crud.find_by_name_or_strain_designation: an exact case-insensitive strain_designation match wins; otherwise a case-insensitive name substring match; otherwise an empty list. Matching strain exactly keeps a short code from polluting name results. Results eager-load the same joins as show().
- Enrich laboratory_person payloads with derived person_display_name, laboratory_name, and laboratory_strain_designation (model properties added to LaboratoryPersonSchemaShow and ...Related, so the two reverse endpoints and the lab_persons nested under Laboratory all carry them).
- Tests for the lookup precedence and the new display fields.

### DIFF
--- a/agr_literature_service/api/crud/laboratory_crud.py
+++ b/agr_literature_service/api/crud/laboratory_crud.py
@@ -1,9 +1,9 @@
 import logging
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from fastapi import HTTPException, status
 from fastapi.encoders import jsonable_encoder
-from sqlalchemy import or_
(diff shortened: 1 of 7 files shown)


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1203, opened 2026-06-18
only auto-create companion entity tag for human-curator tags

sweng66:
The companion pure entity tag was being auto-created for every positive mixed topic+entity tag created via the API, including pipeline/ML-generated tags (they POST through the normal endpoint, so the validate_on_insert gate did not exclude them). Per curator feedback, the auto-creation should mirror manual UI behavior and only fire for tags a human curator made.

- create_tag: gate companion creation on a human creator. Read the final persisted created_by/updated_by off the committed row (the audited-model before_insert copies created_by into updated_by on new_db_obj, not on the request dict) and require both to be AGRKB: curies. Pipeline/script users have non-AGRKB: users.id values, so machine-generated mixed tags no longer spawn companions.
- Back-fill script: add the same created_by/updated_by AGRKB:% filter so historical back-population stays consistent with the live path.
- Tests: model the human path with an AGRKB: curator id on the companion, negated, pure-entity and SGD cases; add test_pipeline_mixed_tag_creates_no_companion; adjust test_validate_negative_with_pos_and_neg count 7 -> 6 (fixture WBPerson1 and service-account default_user are not AGRKB:, so no companion is created).


azurebrd (approved):
Thanks @sweng66

sweng66:
Thanks @azurebrd

PR #1209, opened 2026-06-24
fix(SCRUM-6231): re-enable ksqlDB record cache to cut RocksDB write amplification

valearna:
## Summary

The Debezium → ksqlDB → Elasticsearch reindex was saturating the dev box's EBS volume and slowing the whole instance (the API included).

Diagnosis (SCRUM-6231): the reindex is **disk-write-bound** — the ksqlDB JVM was writing **~405 MB/s** to its local RocksDB state stores, while CPU (~2 of 16 cores) and memory (~50%) had plenty of headroom. Root-cause amplifier: `cache.max.bytes.buffering=0` forced ksqlDB to flush every record straight to RocksDB.

## Change

In `docker-compose.yaml` (`dbz_ksql_server`), re-enable the Kafka Streams record cache and tune for the dev **m5.4xlarge (16 vCPU / 64 GB)**:

- `KSQL_KSQL_STREAMS_CACHE_MAX_BYTES_BUFFERING=104857600` (100 MB) — uses the prefixed `ksql.streams.*` key so it is actually applied (the bare `cache.max.bytes.buffering` was likely ignored). The cache is **per persistent query** (~32 CTAS in `ksql_queries.ksql`), so total on-heap ≈ 3.2 GB.
- `KSQL_KSQL_STREAMS_COMMIT_INTERVAL_MS=10000` — flush every 10 s instead of the 2 s default → larger, fewer writes.
- `KSQL_HEAP_OPTS=-Xmx12g -Xms12g` — up from the 3 GB default to hold the caches (safe on a 64 GB box; Postgres/ES are remote).

## Testing (dev box, sampled during a reindex)

- `%iowait` dropped from **34–53% → 1–2%**
- write latency `w_await` **~2.0–2.7 ms** with a moderate, stable queue (`avgqu-sz` 3.6–8.1) — the EBS volume is keeping up
- batching cut write **IOPS** (larger `avgrq-sz`) so the same byte volume no longer hits the EBS ceiling
- The disk is no longer the bottleneck; load shifted onto the previously-idle CPU cores — the intended trade.

## Related

- **SCRUM-6134** — separate AWS disk for Debezium/Kafka/ksqlDB (the infra half). Note from this investigation: the heavy writer is ksqlDB's RocksDB state dir (`ksql.streams.state.dir`), not Kafka, so a dedicated disk should host that path.

### DIFF
--- a/debezium/setup.sh
+++ b/debezium/setup.sh
@@ -5,19 +5,23 @@ source /status_manager.sh
 
 # Configure sleep timings based on environment
 if [[ "${ENV_STATE}" == "test" ]]; then
-    # Test environment - much shorter sleeps for smaller datasets
+    # Test environment - much shorter sleeps for smaller datasets.
+    # Fixed on purpose: test ignores the DBZ_KSQL_SETUP_SLEEP / DBZ_DATA_PROCESSING_SLEEP env vars,
+    # so test runs stay fast regardless of what is configured for prod.
     CONNECTOR_SETUP_SLEEP=0
     KSQL_SETUP_SLEEP=5
     KSQL_POST_SLEEP=10
     DATA_PROCESSING_SLEEP=20
     echo "Running in TEST mode with reduced sleep timings"
 else
-    # Production environment - longer sleeps for full datasets
+    # Production environment - longer sleeps for full datasets.
+    # The two big waits are overridable via the DBZ_-prefixed env vars (wired through
+    # docker-compose), falling back to the safe production defaults when unset.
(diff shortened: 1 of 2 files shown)


sweng66 (dismissed):
Thanks for the fixes. @valearna

valearna:
thanks @sweng66, can you re-approve this PR? I've added env vars to control the sleep time (the dev server is probably much faster than prod now that it is on GP3)

sweng66 (approved):

PR #1210, opened 2026-06-24
add batch /by_references endpoint to speed up the Topic grid

sweng66:
The TET validation (Topic) grid fetched tags with **one HTTP call per reference**
(`GET /topic_entity_tag/by_reference/{curie}`) — the grid's main performance bottleneck on a full search page. This adds a batch endpoint so the grid can fetch every reference's tags in a single request.

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -15,7 +15,7 @@ from fastapi import HTTPException, status
 from fastapi.encoders import jsonable_encoder
 from sqlalchemy import case, and_, create_engine, text
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import Session, joinedload, sessionmaker, noload
+from sqlalchemy.orm import Session, joinedload, selectinload, sessionmaker, noload
 
 from agr_literature_service.api.crud.topic_entity_tag_utils import get_reference_id_from_curie_or_id, \
     get_source_from_db, add_source_obj_to_db_session, get_sorted_column_values, \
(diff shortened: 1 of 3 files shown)


sweng66:
@claude Can you check again?

valearna:
Thanks @sweng66! Not sure I understand how the new endpoint works @sweng66. Is it a single query to the db or multiple queries but this is still faster than single endpoints because it removes the network overhead?

sweng66:
Fixes by Claude + suggested fixes from codex. @valearna 

From Claude:

It's multiple DB queries, not a single one

show_all_reference_tags_for_references does, per batch request:
1. Resolve each identifier → reference_id (a query per id).
2. One union query to build the curie→name map across all references' tags.
3. Then it loops show_all_reference_tags per reference, and each iteration runs several queries: the tags query (+ a selectinload query for validated_by), a user-display-name query, and a db.query(ModModel).all().

So total DB work is O(N) queries (roughly a handful per reference) — definitely not one query.

Why it's still much faster — but it's two network layers, not one

Your colleague is right that network overhead is the win, but there are two layers of it removed, plus an N+1:

1. Browser → API round-trips (the obvious one): N separate HTTP requests (each with TLS/latency/auth, browser-capped at ~10 concurrent) collapse into 1 batched POST. This is the headline win.
2. API → external A-team name lookups: this was the sneakier cost — the per-reference path made ~3 external HTTP calls per reference. The batch resolves names once across the union (~3 total). That's another whole layer of network round-trips eliminated, server-side.
3. The validated_by N+1 (a DB issue, not network): each tag was lazy-loading its validators in a separate query; now eager-loaded once.

So: multiple DB queries, but the speedup comes from removing two layers of network round-trips (and one N+1) — not from reducing DB query count. The local DB queries were never the bottleneck; the network hops (browser↔API, and especially API↔A-team) were.

The remaining ceiling

If you wanted it to also be (close to) a single DB query, that's the deferred optimization #2 from the review: replace the per-reference loop with one tags query across all reference IDs + one shared user map + one mod map. That trims the O(N) local queries down to ~constant — worth doing only if a full 100/page still lags, since local DB queries are cheap relative to the network hops already eliminated.

valearna:
cool, makes sense to keep one query per reference, at least for now. Thanks @sweng66 !

valearna (approved):

sweng66:
Thanks @valearna

PR #1211, opened 2026-06-25
perf(debezium): ksqlDB reindex — setup-race fix + RocksDB compression + event-driven gating (SCRUM-6231)

valearna:
## Summary

Improves the Debezium → ksqlDB → Elasticsearch reindex along three axes:
1. **Correctness** — fix the ksql-setup race that produced an empty index.
2. **Disk** — cut the RocksDB write volume that saturated the dev instance (compression), instead of throttling.
3. **Idle time** — replace the two blind fixed `sleep`s in the reindex with event-driven gating.

All on branch `SCRUM-6231`.

---

## 1. Fix: ksql-setup batch race → empty index (correctness)

`debezium/setup.sh` now submits ksql statements **one per request** (`submit_ksql_statements`) instead of one big batch. The batch POST hit a metastore-visibility race: a CTAS referencing a table created earlier in the *same* batch failed with `"<TABLE> does not exist"`, which aborted every later statement — including the final `reference_joined` / `public_reference_joined` tables the ES sinks consume → **0-doc index**. Sequential submission makes each statement commit before the next is prepared; carries `auto.offset.reset=earliest` via `streamsProperties`; idempotent on "already exists" + short dependency retry.

## 2. RocksDB compression — cut disk write volume (the saturation fix)

The reindex was throughput-bound at the m5.4xlarge **~594 MB/s instance EBS ceiling**, dominated by ksqlDB's RocksDB state-store writes. Env-only knobs can't reach RocksDB's on-disk behaviour, so this bakes a config-setter into a custom ksqlDB image:

- **`docker/ksqldb/KsqlRocksDBConfigSetter.java`** (new) — LZ4 + ZSTD-bottommost compression, universal compaction, and one shared auto-tuned writes-only `RateLimiter` (cap = `disk.max × fraction`). Every lever env-tunable via `KSQL_KSQL_STREAMS_ROCKSDB_*` with safe defaults; logs via slf4j.
- **`docker/ksqldb.dockerfile`** (new) — multi-stage build compiling against ksqlDB 0.26's *own* kafka-streams 3.2 / rocksdbjni 6.29.4.1 jars (exact API match), jar onto the classpath.
- **`docker-compose.yaml`** — `dbz_ksql_server` gets `build:` + `image: agr.literature.ksqldb-server:0.26.0` + `KSQL_KSQL_STREAMS_ROCKSDB_CONFIG_SETTER`. ⚠️ The Streams key is `rocksdb.config.setter` (**not** `…class`) — the `_CLASS` form is silently ignored (`rocksdb.config.setter = null`). Also includes the env tuning that helps but can't reach RocksDB write *volume*: record cache 150 MB, heap 16 g, `topology.optimization=all`, producer `zstd`, `standby.replicas=0`.
- **`Makefile`** + **`.github/workflows/debezium-integration-test.yaml`** — `up -d --build dbz_ksql_server` so the custom image is rebuilt/redeployed by `make restart-debezium-*` and exercised in CI.

**Measured on the dev box (A/B vs the no-compression baseline):** sustained write **−32%** (411→278 MB/s), peak **−34%** (648→428), write IOPS **~−53%**, on-disk state **~−49%** (139→71 GB), queue depth **−56%**, w_await **−39%**, iowait **−63%** — and it finished **~40 min faster** with both indexes complete (**no throttling penalty**). Verified on-box via the RocksDB OPTIONS file (`compression=kLZ4Compression`, `bottommost=kZSTD`).

> For the record: `num.stream.threads=2` was tried as an env-only throttle and **reverted** — the A/B showed it *tripled* the run while still saturating the disk. Compaction write-amplification is driven by data volume + LSM structure, not thread count.

## 3. Event-driven reindex gating — replace fixed sleeps

`debezium/setup.sh` + `debezium/status_manager.sh` replace the two blind `sleep`s in the **production** path with condition polling (HTTP-only; the `dbz_setup` container has just curl/jq). `DBZ_KSQL_SETUP_SLEEP` / `DBZ_DATA_PROCESSING_SLEEP` are kept as **max-wait caps** (hard fallback), so it's backward-compatible and adapts to actual duration on any instance.

- **Gate 1 (`wait_for_source_topics_ready`)** — submit ksql once the Debezium snapshot has created the source topics (Kafka Connect topic tracking; proceed at ≥ `table.include.list` count with the connector RUNNING). ksql's `earliest` offset reset consumes the full snapshot as it streams in, so topic-exists is sufficient.
- **Gate 2 (`wait_for_pipeline_drained`)** — promote temp→final once **both** temp indexes' **`docs.count` + `store.size_in_bytes` + `index_total`** are strictly unchanged for **5 min** (`DBZ_DRAIN_STABLE_SECONDS=300`, poll 30 s) **and** both ES sink connectors are RUNNING with no FAILED tasks **and** both indexes are non-empty. Doc count alone is invalid (the sink upserts by id, so overwrites keep count flat) — `index_total`/`store.size` catch that activity. Strict no-change is intentional; the cap is the fallback if live CDC keeps it moving.

**Test mode is unchanged** (keeps the fixed short sleeps + existing index-count poll), so the integration test's behavior/timing is preserved.

---

## Files changed
| File | Change |
|---|---|
| `docker/ksqldb/KsqlRocksDBConfigSetter.java` (new) | RocksDB config-setter (compression + universal compaction + rate limiter) |
| `docker/ksqldb.dockerfile` (new) | Multi-stage custom ksqlDB image |
| `docker-compose.yaml` | ksqlDB tuning + config-setter wiring + custom image build |
| `debezium/setup.sh` | Sequential ksql submission + event-driven Gate 1/Gate 2 + configurable sleeps |
| `debezium/status_manager.sh` | `wait_for_source_topics_ready`, `wait_for_pipeline_drained` |
| `Makefile` | `--build dbz_ksql_server` on restart/test targets |
| `.github/workflows/debezium-integration-test.yaml` | `up -d --build dbz_ksql_server` |

## New env vars
- `DBZ_KSQL_SETUP_SLEEP` (1200) / `DBZ_DATA_PROCESSING_SLEEP` (20000) — now max-wait caps.
- `DBZ_SNAPSHOT_POLL_INTERVAL` / `DBZ_SNAPSHOT_STABLE_POLLS`; `DBZ_DRAIN_POLL_INTERVAL` / `DBZ_DRAIN_STABLE_SECONDS`.
- `KSQL_KSQL_STREAMS_ROCKSDB_CONFIG_SETTER` + `KSQL_KSQL_STREAMS_ROCKSDB_*` tuning (compression/compaction/ratelimit), all with in-code defaults.

## Verification
- **CI**: `run-debezium-integration-tests` green (exercises ksql setup + a live CDC sync end-to-end); flake8 / mypy / unittest / functest / CodeQL green.
- **Gate logic**: local stubbed-curl unit tests — strict-stability→drain, failed-sink→cap, never-stable→cap, gate1→ready.
- **Compression + config-setter**: verified on the dev box via the RocksDB OPTIONS file + the A/B disk profile above.

## Out of scope / follow-up
- **SCRUM-6134** (instance-store NVMe) remains the lever for *full* disk headroom — compression got the disk meaningfully off saturation, but the core build still runs ~85% util. Compression stacks under it.


valearna:
@claude check again

sweng66 (approved):
Thanks for working on this. @valearna

valearna:
thanks @sweng66 !

PR #1212, opened 2026-06-26
feat(api): load resource descriptors from A-Team with YAML fallback (SCRUM-5992)

ianlongden:
## Summary
A-Team is becoming the source of truth for resource descriptors. This loads them from the A-Team curation API at startup / `PUT /resource_descriptor/` reload, falling back to the existing agr_schemas YAML during the transition.

- Refactors `api/initialize.py`:
  - `_fetch_descriptors_from_ateam()` — calls `AGRCurationAPIClient().get_resource_descriptors()` and normalizes A-Team fields → ResourceDescriptor model.
  - `_fetch_descriptors_from_yaml()` — existing GitHub YAML path (unchanged behavior).
  - `_load_descriptors_into_db()` — extracted DB loader shared by both sources.
  - `update_resource_descriptor()` — tries A-Team first, falls back to YAML if it errors or returns nothing.

## Field mapping (confirmed vs agr_curation source)
`prefix→db_prefix`, `name→name`, `synonyms→aliases`, `idExample→example_gid`, `idPattern→gid_pattern`, `defaultUrlTemplate→default_url`, `resourcePages[{name, urlTemplate}]→pages[{name, url}]`.

## Tests
`tests/api/test_resource_descriptor.py`: field-mapping unit test, A-Team DB load, and both fallback paths (A-Team error + empty). flake8 + mypy clean.

## Dependency / rollout note
Requires the `get_resource_descriptors()` method added in alliance-genome/agr_curation_api_client#20. `requirements.txt` is unchanged here; until that library change is released and the pin is bumped, the A-Team path raises and **safely falls back to the YAML source** — so this is mergeable independently with no runtime breakage.

### DIFF
--- a/agr_literature_service/api/initialize.py
+++ b/agr_literature_service/api/initialize.py
@@ -4,6 +4,7 @@ import sys
 import urllib.request
 from urllib.error import URLError
 from urllib.parse import urlparse
+from typing import Any, Dict, List, Optional
 import yaml
 import logging
 from sqlalchemy.orm import Session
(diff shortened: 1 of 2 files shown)


ianlongden:
@claude re review code

sweng66 (approved):
Thanks @ianlongden

PR #1213, opened 2026-06-26
speed up Search Topic grid loading

sweng66:
codex did the following fixes to API:
* batch loading optimization
* bulk reference resolver
* curie name-map cache

claude did the following fixes after codex to API:

* get_tet_with_names — copy.deepcopy(tet) is gone (the only remaining match at line 1467 is a code comment explaining the replacement); the actual logic now uses new_tet = dict(tet) (two-level shallow copy, input untouched, same return contract).

* _serialize_reference_tag_rows — jsonable_encoder(vars(tet)) is gone (no grep match); replaced by _orm_column_keys / _project_orm_columns reading mapped scalar columns via getattr, with response_model still doing the final JSON encoding.

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -9,6 +9,7 @@ from collections import defaultdict
 from os import environ
 from typing import Any, Dict, List, Set, Tuple
 from datetime import datetime, timedelta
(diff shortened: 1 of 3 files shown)


sweng66:
@claude Can you check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #1216, opened 2026-06-26
feat(debezium): alias-based blue/green index swap, no reindex (SCRUM-…

valearna:
…6240)

Replace the delete-final-and-reindex-in-place promotion with a zero-downtime blue/green swap:

- The app queries an ALIAS (= DEBEZIUM_INDEX_NAME); physical data lives in two fixed slots <name>_1 / <name>_2.
- Each rebuild targets the INACTIVE slot (the live slot is never touched), then force_merges (optimize + expunge tombstones) and warms it OFF the serving path before an atomic _aliases flip. The previous slot stays as an instant-rollback backup; disk is bounded at two copies.
- No temp->final reindex (dropped): the streamed slot is promoted directly.
- One sink connector per index, retargeted to the inactive slot at build start and kept after the flip -- the cutover is alias-only, decoupled from Kafka Connect (no temp/permanent connector juggling).
- Guard: never flip to an empty slot, so a failed/empty build cannot replace a healthy live index. flip_alias bootstraps off the legacy bare index name.
- Same flow for test and prod (only wait timings differ).

No API change: ELASTICSEARCH_INDEX already equals the alias name, so search resolves through the alias transparently.

The old reindex helpers (start_reindex / poll_reindex_task / poll_multiple_reindex_tasks) are now unused; left in place to keep this diff focused and can be pruned in a follow-up.

### DIFF
--- a/debezium/setup.sh
+++ b/debezium/setup.sh
@@ -245,6 +245,7 @@ if [[ $new_index_doc_count -gt 0 ]] && [[ $public_index_doc_count -gt 0 ]]; then
     else
         echo "ERROR: alias flip was NOT acknowledged; the live index was NOT switched to slot _${SLOT}. The previous slot keeps serving."
         set_reindex_status "error" "{\"message\": \"alias flip failed; live index unchanged\", \"slot\": \"${SLOT}\"}"
+        exit 1
     fi
 else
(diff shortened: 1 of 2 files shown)


valearna:
@claude check again

sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #1217, opened 2026-06-29
feat(pdf2md): persist & expose PDFX figure metadata sidecars + backfill (SCRUM-6246)

valearna:
## What & why

Addresses the ABC Literature side of **KANBAN-1384** (Doug Howe / AI Curation need figure-and-panel provenance plus figure **legends** for gene-expression extractions). Today the pdf2md pipeline fetches the PDFX/Marker image manifest but **discards all of its metadata** — only the figure PNGs are persisted as ordinary `referencefile` rows. This PR persists each figure's manifest metadata as a JSON sidecar so AI Curation can query legends and map evidence to a figure.

Tracked as **SCRUM-6246**.

## Changes

1. **Persist a metadata sidecar** (`lit_processing/pdf2md/pdf2md_utils.py`) — `process_extracted_images` now uploads a deterministic JSON sidecar per figure: same `display_name` as the figure PNG, `.json` extension, under new file_classes `converted_main_figure_metadata` / `converted_supplement_figure_metadata`. Fields persisted: `figure_label`, `figure_number`, `caption_text` (the legend), `nearby_text`, `page_index`, `bbox`, `polygon`, `filename`, and `image_review_*`. Serialized with sorted keys → md5sum-stable → idempotent; the transient presigned `url` is excluded. Sidecar failures are logged but never fail the conversion. This single change covers **both** the cron batch (`pdf2md.py`) and the on-demand conversion API (both funnel through `_process_single_pdf_file`).
2. **Expose it** (`api/schemas/file_conversion_schemas.py`, `api/crud/file_conversion_crud.py`) — `ConversionFileInfo` gains `metadata_referencefile_id`, and `_figures_for_source` links each figure PNG to its sidecar (matched by shared `display_name`). `show_all` returns the sidecar rows unchanged. **No DB column / migration** — metadata lives in the JSON referencefile.
3. **Backfill** (`lit_processing/pdf2md/backfill_figure_metadata.py`, new) — finds already-converted references whose figures lack a sidecar and re-runs each source PDF through **marker-only** PDFX (`methods=marker, merge=False` — skips Grobid/Docling/merge to **save credits**; image-review left on). Attach-only (never re-uploads PNGs), idempotent, resumable, `--dry-run`, single-`--reference`/`--curie` mode, `--mod`, `--limit`, `--no-supplements`. Targets **only** the Marker `converted_*_figure` classes and refuses a source whose figure names are non-canonical or whose regenerated manifest count doesn't match, to avoid mis-associating a caption to the wrong figure.

## Scope (verified against prod)

276 references / ~2,786 Marker-extracted figures need backfill (274 still have a source PDF). This explicitly **excludes** the ~3.14M-file `figure` class (images downloaded from PMC/PubMed or manually uploaded) — those are never touched.

## Testing

Unit tests added/updated for the payload builder, deterministic serialization, the sidecar upload (incl. failure isolation and the PNG-rename case), the API linking, and all backfill planning branches. Verified via CI (the test DB is wiped if run locally). `flake8` + `mypy` pass.

## Independent review

An independent review confirmed all 4 acceptance criteria are met (no blocker/high). Two MEDIUM findings were fixed in this branch: `figure_index` is now parsed from the PNG ordinal so it's consistent between the live and backfill paths; and the backfill refuses sources with non-canonical figure names (plus a per-attach audit log).

---

## @christabone — consumer-contract questions (please review)

The metadata is **one JSON sidecar file per figure** (not a separate store/columns) — every field, **including `image_review_*`, lives inside that single `.json`**. Could you confirm the contract so AI Curation can build against it without rework?

- **`image_review_*`** — these are PDFX's per-image LLM image-review fields. I persist them **inside the same sidecar, nested under an `image_review` object** (original keys preserved, e.g. `image_review.image_review_score`). Do you want them **flat at the top level** instead (`image_review_score`, …)? Easy to change.
- **PNG → metadata mapping** — will you map via the **shared `display_name`** (through `show_all`), via **`metadata_referencefile_id`** (conversion API figures listing), or both?
- **Legend field** — is **`caption_text`** the legend you need, or should you read `nearby_text`? (Both are persisted.)
- **Panel-level provenance** — the PDFX manifest is **figure-level** (no panel field), so this delivers figure number + legend + bbox but **not panel numbering**. Is figure-level sufficient for the gene-expression use case, or do you need panels (which would be a separate, PDFX/extraction-side problem)?
- **`bbox`/`polygon`** — do you need page dimensions to interpret the coordinates, or is the raw bbox/polygon enough? (`page_index` is included; page size is not.)

Coordination (non-blocking): the backfill re-runs marker-only PDFX on ~274 references (credits + time) — happy to align on timing, and to ship the live-path change first so no new gaps accumulate during the backfill.

### DIFF
--- a/agr_literature_service/api/crud/file_conversion_crud.py
+++ b/agr_literature_service/api/crud/file_conversion_crud.py
@@ -454,6 +454,13 @@ _FIGURE_FILE_CLASS_FOR_SOURCE: Dict[str, str] = {
     "supplement": "converted_supplement_figure",
 }
 
+# Parallel mapping for the JSON metadata sidecar each figure now carries
+# (SCRUM-6246). Kept in sync with FIGURE_METADATA_FILE_CLASSES in pdf2md_utils.
+_FIGURE_METADATA_FILE_CLASS_FOR_SOURCE: Dict[str, str] = {
+    "main": "converted_main_figure_metadata",
+    "supplement": "converted_supplement_figure_metadata",
+}
+
 
 def _figures_for_source(reference: ReferenceModel,
                         source_display_name: Optional[str],
@@ -465,13 +472,40 @@ def _figures_for_source(reference: ReferenceModel,
     ``{source_display_name}_image_{idx:03d}`` convention and one of the
     ``converted_main_figure`` / ``converted_supplement_figure`` file_classes,
     so we match on display_name prefix + the corresponding figure file_class.
+
+    Each figure's JSON metadata sidecar (SCRUM-6246) shares the PNG's
+    display_name under the parallel ``converted_*_figure_metadata`` file_class
+    (``json`` extension), so we link the two by exact display_name match and
+    expose the sidecar as ``metadata_referencefile_id``.
     """
     if not source_display_name or not source_file_class:
         return []
     figure_file_class = _FIGURE_FILE_CLASS_FOR_SOURCE.get(source_file_class)
     if figure_file_class is None:
         return []
+    metadata_file_class = _FIGURE_METADATA_FILE_CLASS_FOR_SOURCE.get(
+        source_file_class
+    )
     prefix = f"{source_display_name}_image_"
+
+    # display_name -> metadata sidecar referencefile_id, so each figure PNG
+    # can point at its sidecar without re-querying the DB per figure.
+    metadata_by_display_name: Dict[str, int] = {}
+    if metadata_file_class is not None:
+        for ref_file in reference.referencefiles or []:
+            if ref_file.file_class != metadata_file_class:
+                continue
+            if ref_file.file_extension != "json":
+                continue
+            if ref_file.file_publication_status != "final":
+                continue
+            display_name = ref_file.display_name or ""
+            if not display_name.startswith(prefix):
+                continue
+            metadata_by_display_name[display_name] = int(
+                ref_file.referencefile_id
+            )
+
     figures: List[Dict[str, Any]] = []
     for ref_file in reference.referencefiles or []:
         if ref_file.file_class != figure_file_class:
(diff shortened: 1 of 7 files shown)


christabone:
Thanks, Valerio. From the AI Curation side, this contract looks good as-is; I do not think any code changes are needed for these questions.

Answers inline:

- `image_review_*`: nested under `image_review` is preferred. That keeps the sidecar tidy and makes these fields clearly auxiliary PDFX/image-review metadata rather than top-level figure identity fields.
- PNG to metadata mapping: AI Curation should use `metadata_referencefile_id` when consuming the conversion API figures listing. For broader `show_all` discovery/backfill/general referencefile listing, the shared `display_name` + `.json` sidecar remains the fallback/general mapping contract. Supporting both is useful.
- Legend field: `caption_text` is the primary figure legend field we should build against. `nearby_text` is useful supporting context/fallback, but should not be treated as the canonical legend when `caption_text` is present.
- Panel-level provenance: figure-level metadata is sufficient for this ABC Literature-side PR. Panel-level provenance remains a separate AI Curation/text-evidence extraction problem unless PDFX later exposes panel-aware metadata. We should not block this PR on panel support.
- `bbox`/`polygon`: raw bbox/polygon plus `page_index` is enough for the initial AI Curation use case. Page dimensions may become useful later if we build visual overlays or coordinate transforms, but they are not required for the current figure legend/provenance contract.

For rollout, I agree with shipping the live-path sidecar persistence first so no new gaps accumulate, then coordinating the marker-only backfill separately after a dry run / small single-reference run confirms the counts and cost envelope.

christabone (approved):

PR #1219, opened 2026-06-30
fix(pdf2md): suppress file-upload WFT transition during figure-metadata backfill (SCRUM-6246)

valearna:
## Summary

Follow-up fix to #1217 (SCRUM-6246). Surfaced while running the figure-metadata backfill on **stage** (one main + one supplement reference).

### The bug
The backfill uploads one JSON metadata sidecar per converted figure through the shared `file_upload()` path, which fires `transition_WFT_for_uploaded_file`. For already-converted references the ATP descendants cache is **cold in the one-off backfill process**, so `get_current_workflow_status(...)` returns `None` on every call. The first sidecar then *creates* the "file upload in progress" tag (`ATP:0000139`) and the **2nd+ sidecar of the same reference collides on the unique file-upload WFT → 422**. It also spuriously regresses a reference that was already at "files uploaded" (`ATP:0000134`).

Observed on stage: sidecar #1 attached, #2–#8 failed with `422 ... ATP:0000139 already exist ... can not create duplicate record`.

### The fix
A **process-level toggle** in `referencefile_crud` (mirrors the transient SCRUM-5795 `converted_*` upload bypass that SCRUM-5867 removed):

- `set_suppress_post_upload_workflow(True/False)` + `is_post_upload_workflow_suppressed()`
- When active, `file_upload()` still persists **and commits** the uploaded file but skips `transition_WFT_for_uploaded_file` and `cleanup_old_pdf_file`.
- The **backfill activates it for the duration of each run** (reset in `finally`). It's a standalone process, so the global never leaks into the API/cron conversion processes.
- **Live API / conversion path is unchanged** (default off) — real conversion artifacts still drive the workflow.

Rationale: these references were already converted; attaching auxiliary descriptor files must not move the file-upload workflow status or trigger downstream reprocessing.

### Verification (stage, against `literature-dev`)
- Cleaned the stray `ATP:0000139` left by the buggy run on `AGRKB:101000000657446`; confirmed it returns to `ATP:0000134` with 8 PNGs ↔ 8 sidecars.
- Ran the fixed backfill on a **pristine** supplement reference `AGRKB:101000001213127`: **6 sidecars added, 0 errors**, 6 PNGs ↔ 6 sidecars, **no stray `ATP:0000139`**, workflow stays at `ATP:0000134`. Sidecar JSON well-formed (manifest fields + nested `image_review`, `url` excluded).
- `flake8` + `mypy` clean; added unit tests for the toggle in `tests/api/test_file_conversion.py`.

### DIFF
--- a/agr_literature_service/api/crud/referencefile_crud.py
+++ b/agr_literature_service/api/crud/referencefile_crud.py
@@ -49,6 +49,31 @@ file_needed_tag_atp_id = "ATP:0000141"
 text_conversion_process_atp_id = "ATP:0000161"
 
 
+# TRANSIENT (SCRUM-6246 figure-metadata backfill): a one-off process may flip
+# this so file_upload() persists *auxiliary* files (e.g. the figure-metadata
+# JSON sidecars) WITHOUT firing transition_WFT_for_uploaded_file or pruning
+# superseded main PDFs via cleanup_old_pdf_file. The reference's file-upload
+# workflow status was already set when the figures were originally converted, so
+# re-attaching descriptor files must not move it or trigger downstream
+# reprocessing. Default is False, so the live API / conversion path is
+# unaffected — only the backfill script activates it, for the duration of its
+# run (see lit_processing/pdf2md/backfill_figure_metadata.py). Process-level
+# (module global): the standalone backfill runs in its own process, so this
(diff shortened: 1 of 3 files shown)


sweng66 (approved):
Thanks @valearna

PR #1220, opened 2026-06-30
feat(embedding): embedding_file catalog table + referencefile API + register endpoint (SCRUM-6141)

valearna:
**Draft** — will be marked ready for review after today's prod deployment.

## Summary (SCRUM-6141 — "create table in ABC to store embedding metadata", epic SCRUM-6139)

Adds a lean `embedding_file` catalog that maps each stored embedding parquet to its reference + recipe, reusing the existing `referencefile` storage. **The parquet *is* a `referencefile`** (`file_class='embedding'`); `embedding_file` is a non-audited sidecar holding the semantics + two FKs into `referencefile`.

### Schema
`embedding_file` keyed by `(reference_id, profile_name, version, source_referencefile_id)`:
- `source_referencefile_id` → the `converted_merged_*` markdown that was embedded (`NULL` for abstracts)
- `parquet_referencefile_id` → the stored parquet referencefile
- `profile_name`, `version`, `model_name`
- No vectors / recipe / md5 here (those live in the parquet + on the parquet referencefile row).
- PG13 has no `NULLS NOT DISTINCT`, so the unique key is enforced via **two partial unique indexes** (the `referencefile_mod` pattern).
- Migration `1a1274ba1f52` (single alembic head; chains off current main head).

### Retrieval / download (reuse existing referencefile APIs)
- `GET .../referencefile/show_all/{curie}?include_embeddings=true&profile_name=&version=` — embedding rows carry `profile_name`/`version`/`model_name` + an upward `source` sub-object; default-excluded otherwise (backward compatible).
- `conversion_request` gains an `embeddings` list per converted file.
- Download = existing `download_file/{referencefile_id}` (parquet is a referencefile) — unchanged.

### Write / lookup (new dedicated router — added per request)
- `POST /reference/embedding_file/` — upload the parquet + upsert the catalog row (idempotent on the unique key); wraps `embedding_file_crud.create_or_update`, which stores the parquet via `file_upload_single` (no file-upload workflow side effects).
- `GET /reference/embedding_file/{id}` — fetch a catalog row.
- Added a `reference_curie` accessor on the model so rows serialize into `EmbeddingFileSchemaShow`.

### Deletion (no new endpoint — by design)
Deleting the parquet `referencefile` (`DELETE .../referencefile/{id}`, ALL_ACCESS) **cascades** the `embedding_file` row away (`ON DELETE CASCADE`) and cleans up S3. No catalog `DELETE` endpoint, since deleting only the row would orphan the parquet. (Note: deleting the source markdown also cascades the row but leaves the parquet — relevant on re-conversion.)

### Tests
`tests/api/test_embedding_file.py`: model columns / partial unique indexes, schema shapes, idempotent `create_or_update`, `show_all` exclude-by-default vs `?include_embeddings=true` (+ `source` lineage), `conversion_request` embeddings attach, and the new `POST`/`GET` endpoints. flake8 + mypy clean. (Test suite runs on CI.)

### Out of scope
Embedding *generation* workflow (SCRUM-6142); optional `embedding_dim`/`chunk_count` columns (deferred pending Chris — purely additive).


valearna:
@claude check again

valearna:
**Status: this PR is ready for review.** I'll move it out of draft after today's prod deployment.

## Design decision: the create endpoint keeps the parquet-upload model

We considered having `POST /reference/embedding_file/` accept the raw embedding **fields** (vectors + per-chunk metadata) and **construct** the parquet server-side in a fixed layout, so the ABC would own/enforce the canonical format. After weighing it, **we're keeping the current upload model** — the caller builds the parquet and uploads it; the endpoint stores it (as an `embedding` referencefile) and upserts the catalog row. Rationale:

- **Payload efficiency** — embeddings are large binary data. A pre-built parquet upload is compact, whereas sending vectors as JSON bloats them ~3–4× and forces the API to assemble parquets in memory.
- **No new runtime dependency** — server-side construction would add `pyarrow` to the API; the producer already has it (and the vectors in memory), so it's the natural place to write the file.
- **Format ownership** — the canonical parquet format is already decided in **SCRUM-6137** (Chris Tabone's embedding-profile spec). Producers can emit it consistently without the ABC re-implementing it.

**Standardization stays achievable later, additively**, without server-side construction:
1. a shared, ABC-owned writer helper that producers call (one definition of the format), and/or
2. upload-time validation in `create_or_update` (reject parquets whose schema doesn't match the spec).

The deferred `embedding_dim` / `chunk_count` catalog columns also remain purely additive if they're ever needed as queryable fields.

## Scope of this PR

Catalog table (`embedding_file`) + migration, the read/discovery surface (`show_all` always lists embeddings; `conversion_request` attaches them per converted file; download via the existing `download_file/{id}`), and the register/lookup endpoints (`POST`/`GET /reference/embedding_file/`). Deletion is handled by deleting the parquet referencefile (cascade), with source-markdown deletion also cleaning up dependent embeddings.

## Next steps — separate stories (epic SCRUM-6139)

The rest of the pipeline is tracked separately:

- **SCRUM-6142 — Add embedding of MD files to file conversion:** the conversion pipeline generates embeddings from the converted markdown and registers them via this endpoint/CRUD (the main producer path).
- **SCRUM-6140 — Embed reference abstracts with a small model:** abstract-level embeddings feeding the abstract classifiers.
- **SCRUM-6143 / SCRUM-6144 — Embed the WB / FB training-set references** (for the ML document classifiers).
- **SCRUM-6145 — Embed older already-converted files for classification** (backfill of the existing corpus).
- **Consumer integrations (to be tracked, partly in other repos):**
  - ML document classifiers (`agr_automated_information_extraction`) switch from on-the-fly averaged BioWordVec vectors to the precomputed ABC embeddings.
  - Biocurator assistant / RAG (`agr_ai_curation`) consume ABC embeddings instead of re-embedding on ingest.
  - (plus any other downstream consumers we identify later)
- **Optional ABC follow-ups:** enforce the standardized format (shared writer helper or upload-time validation once we want it); add `embedding_dim` / `chunk_count` columns if Chris needs them queryable.

christabone:
Thanks Valerio, this looks like a really good foundation for SCRUM-6141. I like the shape of keeping the parquet as a normal `referencefile` and using `embedding_file` as the semantic/catalog sidecar. From the `agr_ai_curation` side, I think there are two contract questions worth clarifying now so we do not make the producer/consumer assumptions diverge later.

1. Should embedding parquet access always inherit from the source artifact?

For full-text embeddings, our expectation is that the embedding parquet is derived content from the converted Markdown/PDF, and therefore should not be accessible more broadly than the source artifact it was generated from. Right now the create endpoint accepts `mod_abbreviation` independently when registering the embedding parquet. That may be fine if callers are expected to pass exactly the same access scope as the source, but from the consumer/security side it feels easy for a producer to accidentally register a restricted-paper embedding under broader access, especially if `mod_abbreviation=None` means all/global access in the normal referencefile path.

The thing we want to avoid is: source Markdown/PDF is restricted to one MOD or access group, but the derived embedding parquet is downloadable under a broader `referencefile_mods` policy. Since the parquet may contain enough chunk text and vector-derived signal to function as a derived copy of the paper, `agr_ai_curation` would need to treat its access policy as at least as sensitive as the source text.

Would you expect the endpoint/CRUD to enforce inheritance from `source_referencefile_id` for full-text embeddings? Or is the intended contract that the producer must pass the correct `mod_abbreviation`, with tests/producer code guaranteeing it? For abstract-level embeddings where `source_referencefile_id` is `NULL`, should access inherit from the reference/corpus/abstract policy somehow, or is that intentionally a separate all-public path?

2. What should `agr_ai_curation` treat as the stable embedding compatibility/discovery contract?

For our downstream use case, we need to decide whether an ABC embedding artifact is compatible with the curation assistant before we download/load it into Weaviate. The minimum things we care about are the profile identity, model, vector dimension/dtype, chunking recipe/source text kind, schema version, and some staleness signal tying the parquet back to the exact source Markdown that was embedded.

I see the catalog currently exposes `profile_name`, `version`, and `model_name`, with the richer recipe/schema details deferred to the parquet metadata/spec. That may be the right split, especially for SCRUM-6141, but I want to make sure we are aligned before `agr_ai_curation` builds against it. Should we treat `profile_name + version + model_name` as the queryable catalog contract, and then validate everything else after downloading the parquet? Or do you expect additive queryable fields later, such as `schema_version`, `embedding_dim`, `chunk_count`, `source_text_kind`, or `source_md5`, so consumers can choose/reject candidates without fetching every parquet?

Relatedly, for the curation assistant profile, I think we will want a clearly named profile like `curation_assistant_v1` with a stable parquet schema and golden fixture before we rely on these artifacts in production. If that belongs in SCRUM-6142 or a separate downstream ticket, that is fine; I mostly want to confirm where the producer/consumer boundary is meant to sit.

Posted by Codex

valearna:
@claude check again

valearna:
Thanks Chris — both are exactly the right questions to pin down now. Here's where things landed (implemented on the branch as of `a93e13e7`, CI green).

## 1. Access: inherited from the source and enforced — never caller-chosen

We went further than "the producer must pass the correct scope." After we talked, the create endpoint is gone entirely:

- **No HTTP create endpoint.** `POST /reference/embedding_file/` was removed; embedding creation is ABC-internal only (`embedding_file_crud.create_or_update`, called by the producers — e.g. the SCRUM-6142 conversion pipeline). The router keeps only `GET /{id}`; discovery/download stay on the existing referencefile APIs.
- **`mod_abbreviation` is gone from the create payload.** `create_or_update` syncs the parquet's `referencefile_mod` rows to *exactly* match the source referencefile's: MOD-restricted source → same MODs; open/PMC source (NULL mod row) → open. A producer cannot register a restricted paper's embedding under broader access, accidentally or otherwise.
- **Abstract embeddings** (`source_referencefile_id` NULL) are open access — the abstract is public reference metadata in ABC, so there's no restricted source artifact to inherit from. If you'd rather tie these to a reference/corpus-level policy, that's an easy follow-up; open seemed right for now.
- **Access changes propagate immediately.** Every `referencefile_mod` mutation path (create / patch / destroy / referencefile merge) re-syncs derived parquets, so an embedding tracks its source's access without waiting for re-registration. Deleting a source through any path (including removing its last mod association, or merging it away) also deletes its derived embeddings — no stranded parquets.
- **Direct edits are blocked.** POST/PATCH/DELETE of a `referencefile_mod` targeting an `embedding` file returns 422; `embedding_file_crud` is the single writer of embedding access. So the invariant you asked for — *a derived embedding is never downloadable more broadly than the text it was computed from* — is enforced at the API level, not by producer convention.

One deliberate semantic to be aware of: a MOD-scoped `DELETE /reference/referencefile/{parquet_id}` now also 422s (it previously would have silently narrowed the parquet's access by removing that MOD's association). Embeddings are removed via their source file, or by an ALL_ACCESS whole-parquet delete.

## 2. Discovery contract for `agr_ai_curation`

Yes — treat **`profile_name` + `version` + `model_name`** as the stable queryable catalog contract for SCRUM-6141, and validate the richer recipe details (dim, dtype, chunking recipe, schema version) post-download from the parquet metadata, per your SCRUM-6137 spec. Two things that already help you choose/reject candidates before downloading:

- `show_all` embedding rows carry an upward `source` sub-object with the source markdown's `referencefile_id` and its current `md5sum` — so you can tie an artifact to the exact source text, and get a staleness signal by comparing that md5 against the source md5 the producer records inside the parquet.
- Re-conversion deletes and recreates the markdown referencefile, which cascades the stale catalog rows (and their parquets) away — a listed embedding always points at a live source.

The richer queryable fields (`schema_version`, `embedding_dim`, `chunk_count`, `source_text_kind`, `source_md5`) remain purely additive on the catalog. Happy to add whichever ones `agr_ai_curation` actually needs for pre-download filtering — say the word and they go in a small follow-up migration rather than blocking this PR.

On `curation_assistant_v1`: agreed. Profile names are free-form strings in the catalog, so nothing here blocks it — defining the named profile, its stable parquet schema, and a golden fixture belongs with the producer work (SCRUM-6142 or a dedicated downstream ticket), with your embedding-profile spec as the source of truth. We'll loop you in on that ticket so the producer/consumer boundary is agreed before anything ships.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

christabone:
Thanks, Valerio. This lines up with what we need from SCRUM-6141.

I’m happy with this PR as the ABC-side catalog/storage foundation rather than treating it as the whole end-to-end AI Curation embedding program. The important contract point for this PR was access, and the current branch addresses that cleanly: creation is ABC-internal, access is inherited from the source artifact, direct access edits on embedding parquets are blocked, and stale/source-deleted parquets are cleaned up.

For `agr_ai_curation`, I’m also fine with `profile_name + version + model_name` being the lean queryable catalog contract at this stage, with the richer compatibility details validated from parquet metadata or added later as additive catalog fields if we discover we need pre-download filtering. The remaining AI Curation work should stay in follow-up scope: define/produce `curation_assistant_v1`, settle the parquet schema/golden fixtures, and add the downstream loader/validation path on our side.

So from my side: this fits our use case as the first foundation layer for centralized embeddings, with the expected downstream producer/consumer work still tracked separately.

posted by Codex

valearna:
Thanks @christabone ! Can you approve the PR? I'll apply the alembic changes and merge it to the main branch

christabone (approved):

valearna:
thanks @christabone !

PR #1218, opened 2026-06-29
Speed up Search Validation topic grid load

sweng66:
Summary:

Added API-side filtering for the batch topic-entity-tag endpoint, including support for topic, confidence, source, data novelty, score range, entity_types, and entities.

Added API-side aggregation for grid “mini-row” data per reference/topic/source, including entity counts, score ranges, confidence levels, notes, and evidence grouping.

Added/updated tests for the API batch response and UI hook response shape.

PR #1221, opened 2026-06-30
server-side aggregates and validate write path for the TET grid

sweng66:
Summary:

▎ What & why
▎
▎ Moves the TET validation grid's per-cell logic server-side so the batch endpoint returns pre-computed aggregates instead of the UI re-deriving everything from raw tags. Adds a thin write endpoint so a curator validation creates the tag and returns just the recomputed cell. Follow-up to #1218.
▎
▎ Changes
▎
▎ POST /topic_entity_tag/by_references now returns, alongside tags/counts/entries:
▎ - validation — per (reference, topic) curator-validation aggregate: {state: positive|negative|conflict, positives, negatives, by_curator: [{name, negated, sources:[{method,label}], species}]}. Mirrors the grid's ValidationCell so it renders without raw curator tags. Topics with no curator validation are omitted (client treats absent as "unvalidated").
▎ - filter_flags — per (reference, topic) booleans {has_any, has_y, has_n, has_note} backing the per-cell filter, computed over all tags of the cell.
▎ - discovery — batch-global {topics:[{curie,name}], sources:[{label,method,secondary_data_provider,data_provider}]}: the distinct topic columns (over all tags) and source labels (over non-curator tags) present in the post-filter batch, so the grid builds its column set and source filter without scanning raw tags.
▎
▎ POST /topic_entity_tag/validate (new) — thin write path for the grid's Validation column. Given {reference_curie, topic, mod_abbreviation, negated, note?, species?}, it resolves the per-MOD ABC curator source (get-or-create), creates the topic-level validation tag through the normal create_tag path (so dedup/upsert/validation behave identically), and returns the single recomputed cell {topic_entity_tag_id, upserted, topic, validation, filter_flags} — letting the grid update one cell without re-aggregating the whole batch.
▎
▎ Notes
▎
▎ - Curator vs non-curator scoping is consistent across aggregates: curator submissions surface in validation, never in counts/entries/discovery.sources.
▎ - my_validation_present is intentionally not server-provided yet — it needs a curator-identity model (serialized created_by is a display name); deferred.
▎
▎ Tests
▎
▎ Added coverage for validation, filter_flags, discovery, and /validate (incl. positive→conflict re-validation and curator-source get-or-create). Also fixes a pre-existing test that created curator tags without data_novelty (required, non-null) so it could never have passed.
▎
▎ Files: 3 changed, +447. flake8 clean (pre-existing warnings only). ⚠️ Run make run-test-bash (Docker) to execute the suite — not yet run in CI.

### DIFF
--- a/agr_literature_service/api/crud/topic_entity_tag_crud.py
+++ b/agr_literature_service/api/crud/topic_entity_tag_crud.py
@@ -1260,7 +1260,7 @@ def _ci_in(column, values):
     return func.upper(column).in_(upper_values)
 
 
-def _apply_batch_tag_filters(query, filters: Optional[Dict[str, Any]]):
+def _apply_batch_tag_filters(query, filters: Optional[Dict[str, Any]]):  # noqa: C901
     """Restrict a TET query to the tags the initial search asked for.
 
     Mirrors the TET facet criteria the search UI sends (searchActions.js). The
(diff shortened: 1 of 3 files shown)


sweng66:
@claude Can you check again?

sweng66:
@claude

sweng66:
@claude Can you check again?

sweng66:
@claude Check again?

sweng66:
@claude Check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #1222, opened 2026-07-01
fix(pdf2md): backfill index-based matching + bypass in-progress guard (SCRUM-6246)

valearna:
## Summary

Two follow-up fixes for the figure-metadata backfill (SCRUM-6246), both surfaced while running it on **production** (275 refs; 1,851 sidecars added, but 331 blocked + 23 skipped).

### 1. Index-based sidecar matching (recovers the 23 skips)
`plan_sidecars_for_source` matched figure PNGs to manifest images by **sorted position** and required **equal counts** — so any reference whose figures were **md5-deduped** at upload was skipped as a "count mismatch." Identical repeated images (logos, repeated panels) collapse to a single stored PNG via `file_upload`'s md5 dedup, leaving **gaps** in the `_image_NNN` numbering (e.g. stored 31 of a 34-image manifest, gaps at 15/27/31). Verified on prod: the max stored `_image_NNN` index **equals** the current manifest count, i.e. the extractor is unchanged — it's dedup, not version drift.

Now each PNG is matched to `manifest[NNN-1]` by its `_image_NNN` **index**, tolerating gaps. The source is skipped only when the manifest count `!=` the highest stored index (extraction genuinely changed / top images deduped — ambiguous, so we don't risk mis-association) or a PNG has a non-canonical name.

### 2. Bypass `is_file_upload_blocked` for the backfill (fixes the 331 blocks)
All 331 `sidecar_errors` were `422: entity extraction … is currently in progress` — `file_upload`'s **pre-upload "job in progress" guard** firing while the live curation pipeline ran entity extraction on ~40 actively-curated refs. That guard is meant for interactive uploads of real source files; it's a false positive for attaching auxiliary metadata sidecars, which those jobs don't read.

Generalized the existing SCRUM-6246 toggle `set_suppress_post_upload_workflow` → **`set_suppress_upload_guardrails`**: when active, `file_upload` also **skips `is_file_upload_blocked`** (in addition to the workflow transition + main-PDF cleanup it already skipped). The backfill activates it for its run only; the live API/conversion path is unaffected (default off, process-level global). Mirrors the old SCRUM-5795 `converted_*` bypass (removed in SCRUM-5867).

## Tests
- `test_backfill_figure_metadata`: dedup-gap index matching (image_004 → `manifest[3]`, not positional `manifest[2]`), extraction-differs skip.
- `test_file_conversion`: renamed guardrails toggle tests + new `job in progress` cases — default **blocks** (422, upload not attempted) vs suppressed **bypasses** (upload proceeds, guard not called).
- flake8 + mypy clean.

## Rollout
Deploy to prod (rebuild the `automated_scripts` image), then re-run the backfill — idempotent, so it mops up the ~40 previously-blocked refs and the ~23 dedup-mismatch refs. (Prod PDFX cold-start is a separate issue already flagged to Chris.)

### DIFF
--- a/agr_literature_service/api/crud/referencefile_crud.py
+++ b/agr_literature_service/api/crud/referencefile_crud.py
@@ -51,27 +51,34 @@ text_conversion_process_atp_id = "ATP:0000161"
 
 # TRANSIENT (SCRUM-6246 figure-metadata backfill): a one-off process may flip
 # this so file_upload() persists *auxiliary* files (e.g. the figure-metadata
-# JSON sidecars) WITHOUT firing transition_WFT_for_uploaded_file or pruning
-# superseded main PDFs via cleanup_old_pdf_file. The reference's file-upload
-# workflow status was already set when the figures were originally converted, so
-# re-attaching descriptor files must not move it or trigger downstream
-# reprocessing. Default is False, so the live API / conversion path is
-# unaffected — only the backfill script activates it, for the duration of its
-# run (see lit_processing/pdf2md/backfill_figure_metadata.py). Process-level
-# (module global): the standalone backfill runs in its own process, so this
-# never leaks into the API or cron conversion processes.
-_suppress_post_upload_workflow = False
+# JSON sidecars) without the guardrails meant for interactive/live uploads.
+# When set, file_upload():
+#   - skips the is_file_upload_blocked "job in progress" guard, so a downstream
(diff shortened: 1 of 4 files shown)


sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #1224, opened 2026-07-01
feat(SCRUM-6238): one-shot person_lineage validate with curator overr…

azurebrd:
…ides; editable canonical people; lineage names

person_lineage_submission validate is now one-shot: it accepts optional curator values (person_subject/object_curie_or_id, relationship, start_date, end_date) that steer the canonical via find_or_create, falling back to the submission when omitted. The submission's claim and person-id links are never modified — only status and the person_lineage_id link are set to record the promotion.

person_lineage:
- PATCH may correct a mis-resolved person (person_subject/object_curie_or_id); the submission link is preserved (its name claim is simply resolved to the corrected person). Unique-key collisions surface as 422, self-pairs rejected.
- Show now exposes person_subject_name / person_object_name (from the joined person objects) so the UI can display name + curie.
- destroy resets linked submissions (validated/duplicate -> pending) so they return to the unvalidated pool cleanly when a canonical is deleted (FK already SET NULL).

Tests cover override/one-shot validate, submission-untouched, unresolved->422, delete-resets-linked-submission, person correction, and correction collision->422.

### DIFF
--- a/agr_literature_service/api/crud/person_lineage_crud.py
+++ b/agr_literature_service/api/crud/person_lineage_crud.py
@@ -8,7 +8,7 @@ from sqlalchemy import or_
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session, selectinload
 
-from agr_literature_service.api.models import PersonLineageModel
+from agr_literature_service.api.models import PersonLineageModel, PersonLineageSubmissionModel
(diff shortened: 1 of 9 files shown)


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1225, opened 2026-07-03
feat(embedding): generate classifier embeddings for merged Markdown on conversion (SCRUM-6142)

valearna:
## Summary (SCRUM-6142 — embedding producer, epic SCRUM-6139)

Producer side of the embedding pipeline: after the `pdf2md` conversion job produces a reference's **merged** Markdown, embed it and register the parquet in the `embedding_file` catalog (SCRUM-6141, already merged).

- **`openai_embedder.py`** — `OpenAIEmbedder` implementing the shared `agr_abc_document_parsers.embeddings.Embedder` contract (`text-embedding-3-small`, batched + retry, tiktoken token preflight).
- **`embedding_generation.py`** — `generate_classifier_embeddings_for_reference()` embeds every **merged** Markdown of a reference — `converted_merged_main` + each `converted_merged_supplement`, **not** the per-method grobid/marker/docling outputs — into one `embedding_file` per source (so multiple per reference). Idempotent (skips sources already embedded → re-runs never re-spend OpenAI); per-source isolation; optional whole-document vector. Access is inherited from the source referencefile (SCRUM-6141).
- **`pdf2md.py`** — `_maybe_generate_embeddings` hook after a reference converts successfully (covers both the PDFX and nXML paths); failures are logged and never flip the conversion result.
- Config/infra: optional `OPENAI_API_KEY`; docker-compose passes it to the `api` + `automated_scripts` services; `.gitignore` protects the local `.env.*` secret files; `openai` added to requirements.

### Dormant until deliberately enabled
The embedding stack (`agr_abc_document_parsers.embeddings` + `openai`) is imported **lazily**, and the whole step is a logged no-op when it or `OPENAI_API_KEY` is missing — so the conversion job is unchanged on the current `agr-abc-document-parsers==1.5.1` pin. Activation requires: (1) the shared **[embeddings] release** (linked PR) merged + the pin bumped, (2) `OPENAI_API_KEY` provisioned (done for rds-dev), (3) budget sign-off.

Profile: `classifier_fulltext_paragraph_chunk_refs_excluded_md_cleaned` v1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ tmp
 .vscode/*
 **/*.gz
 .env.unittest
+# Local env files holding secrets (DB creds, OPENAI_API_KEY, etc.) — never commit.
+.env.rdsdev
+.env.rdsprod
+.env.devserver4006
--- a/agr_literature_service/api/config.py
+++ b/agr_literature_service/api/config.py
@@ -40,6 +40,10 @@ class GlobalConfig(BaseSettings):
     ELASTICSEARCH_PORT: str = Field(..., env="ELASTICSEARCH_PORT")
     ELASTICSEARCH_INDEX: str = Field(..., env="ELASTICSEARCH_INDEX")
 
+    # Embedding generation (SCRUM-6142). Optional: when unset, the pdf2md
+    # conversion job skips embedding generation (feature stays dormant).
+    OPENAI_API_KEY: Optional[str] = Field(None, env="OPENAI_API_KEY")
+
 
 # instantiate once for application‑wide use
 config = GlobalConfig()
--- a/agr_literature_service/lit_processing/embedding/__init__.py
+++ b/agr_literature_service/lit_processing/embedding/__init__.py
@@ -1,0 +1,2 @@
+"""ABC-side embedding generation (SCRUM-6142): the producer that fills the
+embedding_file catalog from converted merged Markdown."""
--- a/agr_literature_service/lit_processing/embedding/embedding_generation.py
+++ b/agr_literature_service/lit_processing/embedding/embedding_generation.py
@@ -1,0 +1,193 @@
+"""Generate + register classifier embeddings for a reference's merged Markdown
+(SCRUM-6142).
+
+For each **merged** Markdown of a reference — ``converted_merged_main`` and every
+``converted_merged_supplement`` (NOT the method-specific grobid/marker/docling
+outputs) — chunk it (``paragraph_pack``: paragraph-level, non-overlapping,
+references excluded), embed the chunks with OpenAI, write the canonical parquet,
+and register it via :func:`embedding_file_crud.create_or_update`. So a reference
+yields one embedding file per merged Markdown (main + each supplement).
+
+Idempotent: a source that already has an ``embedding_file`` row for this
+``(profile, version)`` is skipped, so re-running the conversion job never
+re-spends OpenAI. Per-source isolation: a failure on one source is logged and
+counted, never aborting the others.
+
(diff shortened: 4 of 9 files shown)


valearna:
@christabone — two linked PRs for the embedding pipeline, would appreciate your review:

- **Shared contracts:** https://github.com/alliance-genome/agr_abc_document_parsers/pull/1 — the `Chunker`/`Embedder` interfaces, the `Chunk` parquet record + canonical writer, and `ParagraphPackChunker` (the worked example) in `agr_abc_document_parsers`.
- **ABC producer:** https://github.com/alliance-genome/agr_literature_service/pull/1225 — wires it into the `pdf2md` conversion job to embed every merged Markdown and register it in the `embedding_file` catalog.

### What's implemented vs. what we need from you

The embedder in the ABC PR (`OpenAIEmbedder`) + `ParagraphPackChunker` are the implementation for the **ML document-classifier** profile (`classifier_fulltext_paragraph_chunk_refs_excluded_md_cleaned`): paragraph-level, non-overlapping, references excluded.

For the **biocurator-assistant** profile (`curation_assistant_v1`), we'd like **you to provide the matching implementation from the assistant side** — a `Chunker` (your `by_title` recipe) **and** an `Embedder` conforming to these same interfaces — packaged and exported from the `agr_ai_curation` repo so the ABC can **import it and run it** to generate the assistant-profile embeddings. That keeps the byte-for-byte parity contract owned by you (the consumer), exactly as your embedding-profile spec asks, while the ABC stays the central producer that runs each app's recipe on its behalf.

So the division is:
- **ABC (these PRs):** the interfaces, the catalog/parquet plumbing, the pdf2md trigger, and the classifier profile's chunker + embedder.
- **You (`agr_ai_curation`, follow-up):** the `curation_assistant_v1` chunker + embedder implementations against these interfaces, exported as a package, plus the golden fixtures.

Happy to adjust the interface shape if anything doesn't fit the assistant's needs — better to settle it now, before you build against it.

christabone (dismissed):
Thanks Valerio. I am approving this producer-side PR as well. The wiring here looks compatible with the direction we want for AI Curation: ABC as the central producer/store for reusable embedding artifacts, with each consumer owning the profile-specific chunking contract.

For this PR specifically, I understand the implemented path as the classifier profile: ABC runs `ParagraphPackChunker`, uses the OpenAI embedder, writes canonical parquet, and registers one embedding artifact per merged Markdown source. That is distinct from the future AI Curation assistant profile, which should be produced from our `curation_assistant_v1` chunker once we package/export it from `agr_ai_curation`.

Our intended follow-up on the AI Curation side is:

- provide the assistant-owned `curation_assistant_v1` parser/chunker/profile implementation against the shared interfaces;
- ship golden fixtures so ABC and AI Curation can verify byte-for-byte chunk/content parity;
- update AI Curation to download the assistant-profile parquet and insert the precomputed vectors into Weaviate directly, instead of re-embedding locally with Weaviate `text2vec-openai`.

So I do not see this PR as needing to solve the assistant profile itself. The important producer-side requirement is that this pattern can grow from the classifier profile to multiple registered profiles over the same final merged Markdown sources, including `curation_assistant_v1` when our package is ready.

One related note from the interface PR: if ABC already has this generic `OpenAIEmbedder`, we are fine with using it for the assistant profile too, as long as the recipe pins the exact model contract (`text-embedding-3-small`, 1536 dimensions, no dimensions override) and embeds exactly `Chunk.content`. The consumer-specific piece we need to own is the chunker/profile and parity fixtures, not necessarily a duplicate OpenAI client.

Posted by Codex

sweng66 (approved):

PR #1227, opened 2026-07-06
fix(bulk_download): stream external_ids endpoints and return 401 on i…

azurebrd:
…nvalid token (SCRUM-6265)

The /bulk_download/references/external_ids/ endpoint 502'd on production: it ran an unbounded array_agg over the entire reference x cross_reference tables, materialized the whole result twice in memory, and did so from an async handler that blocked the event loop - so as the DB grew it exceeded the worker timeout/memory. Rewrite it (and the identical resources sibling) to read flat rows through a server-side cursor (yield_per), group per curie incrementally, and return a StreamingResponse. Memory stays bounded, bytes flow immediately, and the JSON-array response shape is unchanged.

Separately, an Okta (wrong-issuer) token made PyJWKClient raise a PyJWTError that was not caught (only jose JWTError was), surfacing as a 500 Internal Server Error. Catch PyJWTError in the auth dependency and return a clean 401 instead.

Add tests/api/test_bulk_download.py covering both endpoints and the 401 case.

### DIFF
--- a/agr_literature_service/api/auth.py
+++ b/agr_literature_service/api/auth.py
@@ -21,6 +21,7 @@ from typing import Any, Callable, Dict, List, Optional
 
 from fastapi import Depends, HTTPException, Request
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
+from jwt.exceptions import PyJWTError
 
 from agr_cognito_py import get_cognito_auth, get_cognito_user_swagger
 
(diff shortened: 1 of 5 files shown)


valearna (approved):
Looks good, thanks @azurebrd !

azurebrd:
Thanks @valearna !

PR #1228, opened 2026-07-08
feat(embedding): restrict classifier embeddings to classifier MODs + wire into manual conversion (SCRUM-6273)

valearna:
## Summary

Follow-on to SCRUM-6142 (part of the ABC embedding pipeline, SCRUM-6139).

Classifier embeddings are only useful to MODs that have an ML topic classifier, so embedding every converted reference wasted OpenAI spend on embeddings no classifier would consume. This PR restricts classifier embedding generation to references in the corpus of a classifier MOD, wires embedding generation into the manual conversion path (which previously generated none), and adds a `--limit` flag to the pdf2md script for smoke-testing.

## Changes

- **Classifier-MOD filter** (`embedding_generation.py`): new `ML_CLASSIFIER_MODS = ("WB", "FB")` constant (hardcoded for now, easy to extend or swap for an `ml_model`-table lookup) and a `_reference_has_classifier_mod` check. `generate_classifier_embeddings_for_reference` now returns `{"skipped": "no_classifier_mod"}` for references outside those MODs. The batch pdf2md cron path is covered automatically. The filter lives in the classifier-specific function, so future non-classifier embeddings (separate files) are unaffected.
- **Manual conversion API now generates embeddings**: a new isolated wrapper `maybe_generate_classifier_embeddings` is invoked from `run_conversion_job` (main-success semantics) and the sync nXML branch (`_execute_sync_nxml`). Idempotency prevents double-embedding when the batch and manual paths overlap. `pdf2md._maybe_generate_embeddings` was refactored to delegate to the same wrapper.
- **`--limit N` for pdf2md**: adds `max_references` to `main()` and lets `--limit` cap the default workflow mode (it already applied to `--since`); rejected when combined with `--newest`.

## Testing

- flake8 + mypy clean.
- Unit tests: new `test_skips_when_no_classifier_mod`; e2e webtests updated (positive path adds a WB corpus association; new negative-path test asserts a non-classifier-MOD reference is skipped with no embeddings).
- **Validated end-to-end on dev/stage**: ran `pdf2md --limit 2` against the dev DB. Two FB references converted via PDFX and each registered one `embedding_file` row sourced from its `converted_merged_main` Markdown (profile `classifier_fulltext_paragraph_chunk_refs_excluded_md_cleaned`, v1, `openai:text-embedding-3-small`; 53 and 40 chunks). `Failed: 0`.

## Deploy note

With these changes, once containers are rebuilt with `agr-abc-document-parsers[embeddings]` and `OPENAI_API_KEY` set, embeddings also fire on manual conversions for WB/FB references (previously the manual path generated none).

### DIFF
--- a/agr_literature_service/api/crud/file_conversion_crud.py
+++ b/agr_literature_service/api/crud/file_conversion_crud.py
@@ -674,6 +674,16 @@ def _execute_sync_nxml(db: Session, reference: ReferenceModel,
         reference_curie=reference.curie,
         mod_abbreviation=assessment["mod_abbreviation"],
     )
+    if success:
+        # Generate classifier embeddings for the freshly-converted merged
+        # Markdown. Isolated + idempotent, and skipped for references not in a
+        # classifier MOD's corpus.
+        from agr_literature_service.lit_processing.embedding.embedding_generation import (
+            maybe_generate_classifier_embeddings,
+        )
+        maybe_generate_classifier_embeddings(
+            db, reference.reference_id, reference.curie
+        )
     return success, error
 
 
--- a/agr_literature_service/api/utils/conversion_processor.py
+++ b/agr_literature_service/api/utils/conversion_processor.py
@@ -274,6 +274,16 @@ def run_conversion_job(job_id: str, reference_id: int, reference_curie: str,
                 f"for reference_id={reference_id}"
             )
 
+        # Generate classifier embeddings for the merged Markdown, mirroring the
+        # cron path's main-only success semantics: embed whenever the main
+        # conversion succeeded, even if a supplement failed. Isolated + idempotent
+        # (skips references without a classifier MOD and sources already embedded).
+        if not main_failure:
+            from agr_literature_service.lit_processing.embedding.embedding_generation import (
+                maybe_generate_classifier_embeddings,
+            )
+            maybe_generate_classifier_embeddings(db, reference_id, reference_curie)
+
         conversion_manager.complete_job(
             job_id=job_id,
             success=overall_success,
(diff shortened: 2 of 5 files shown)


sweng66 (dismissed):
Thanks @valearna

valearna:
Thanks for the review. Addressing the three non-blocking points:

**1. Embedding speed (was: synchronous OpenAI latency).** The document vector was a second, separate OpenAI request per source. Fixed in 0f2ab6c1 by appending it to the chunk list before `embed_chunks`, so chunks + document text go out in a single batched request — one OpenAI round trip per source instead of two. Round-trip latency dominates per-source cost, so this is the meaningful lever for the bulk runs this feeds. Note the chunk batch was already batched (≤100 inputs/request), so it was 2 calls/source, not N. For true large-scale throughput the next lever is concurrency *across* references, which is a driver/entry-point concern (there's no bulk-embed entry point yet) and better as its own ticket than bloating this PR. Leaving the inline `_execute_sync_nxml` path synchronous: it's guarded by the classifier-MOD filter (WB/FB nXML manual conversions only) and now ~halved; deferring it to a background task adds a second DB session for marginal benefit on an interactive one-off.

**2. `--limit` caps jobs, not distinct references.** Intended — it's a smoke-test flag and reprocessing a ref is harmless (idempotent). No change.

**3. `corpus.is_(True)` excludes NULL/undecided corpus rows.** Intended and correct: only references confirmed in a WB/FB corpus should be embedded. File-level MOD access (MOD-specific vs PMC/shared `None`) is deliberately irrelevant — the source selection does not filter by access, so a WB/FB-corpus reference is embedded regardless of whether its merged Markdown is MOD-specific or PMC. No change.

sweng66 (approved):

valearna:
thanks @sweng66 !

PR #1229, opened 2026-07-08
fix(person-xref): enforce uniqueness only among non-obsolete rows (SC…

azurebrd:
…RUM-6257)

person_cross_reference used plain, unfiltered UniqueConstraints, so an obsolete (soft-deleted) row blocked re-adding the same curie/prefix. This diverged from the Biblio cross_reference table, which enforces uniqueness via partial unique indexes gated on is_obsolete IS FALSE.

Mirror the Biblio semantics:
- model: replace the two UniqueConstraints with partial unique indexes (curie WHERE is_obsolete IS FALSE; (person_id, curie_prefix) WHERE is_obsolete IS FALSE AND person_id IS NOT NULL)
- migration b7c3e9f1a2d4: drop the constraints, create the partial indexes
- crud: add is_obsolete IS FALSE to the duplicate checks in create_for_person and patch so obsolete rows no longer block inserts/updates
- test: regression covering "obsolete row does not block a new active row"

### DIFF
--- a/agr_literature_service/api/crud/person_cross_reference_crud.py
+++ b/agr_literature_service/api/crud/person_cross_reference_crud.py
@@ -48,10 +48,13 @@ def create_for_person(db: Session, person_id: int, payload: Dict[str, Any]) -> P
 
     curie_prefix = _curie_prefix(curie)
 
-    # curie is globally unique on person_cross_reference (uq_person_xref_curie).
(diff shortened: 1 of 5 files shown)


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1230, opened 2026-07-09
classify PMC thumbnails by file size (SCRUM-6281)

sweng66:
Real PMC thumbnails are not named "thumb" (only a few JoVE/Royal Society files are); PMC ships each figure as a large image plus a small same-named thumbnail, so ~half of the 'figure' rows are actually thumbnails. The only reliable discriminator is file size.

- classify_pmc_file: add size-aware thumbnail detection (gif < 25KB, jpg/jpeg < 15KB) via shared THUMBNAIL_MAX_SIZE_BYTES thresholds; keep the legacy "thumb" name check and fall back to it when size is unknown.
- Thread raw file size through the ingest pipeline: emit it in the PMC download TSV, parse it in load_pmc_metadata, and pass it in the annual PMC package update.
- Add oneoff_scripts/reclassify_pmc_thumbnails.py to back-reclassify existing 'figure' gif/jpg/jpeg rows using their S3 object sizes (dry-run by default).

### DIFF
--- a/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/load_pmc_metadata.py
+++ b/agr_literature_service/lit_processing/data_ingest/pubmed_ingest/load_pmc_metadata.py
@@ -49,9 +49,10 @@ def build_file_root_mappings(input_file):
     return pmcid_to_xml_root, pmcid_to_pdf_roots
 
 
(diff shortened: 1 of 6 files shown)


sweng66:
@claude can you check again?

valearna (approved):
Thanks @sweng66 !

sweng66:
Thanks @valearna

PR #1236, opened 2026-07-13
Indexing Changes

pjhale:
This should be the full list of indexing changes needed for this story.  Created a new branch expecting to need to do a big merge but that does not seem to be the case.


valearna (approved):
LGTM @pjhale

PR #1237, opened 2026-07-13
feat(data_check): list references that failed and were not reset

ianlongden:
## Summary

`check_wft_in_progress.py` (SCRUM-4532) already counts how many workflow tags it reset vs set to failed each week. The counts tell you *how many* failed but not *which* references. This adds a dedicated per-MOD report section listing the papers that ended up **failed and were not reset this run** — i.e. newly transitioned in-progress → failed because the upload (`start of progress`) is > 6 weeks old and is therefore too old to keep retrying — so curators have concrete curies to follow up on.

- New `failed_refs` structure (`mod_id → phase → [curie, ...]`) threaded through `process_no_parent` / `process_parent`, populated at the exact points that already increment the `failed` counters (so scope = newly-failed only; the `if not failed:` guard keeps already-failed items out).
- New `get_curie()` helper extracts the curie from the row already fetched for the existing Slack message (falls back to reference_id) — no extra DB queries.
- `report_counts()` emits a labelled section to stdout and appends it to each MOD's `{mod}_check_workflow_problems.log`:

```
[WB] Failed & NOT reset this run (upload > 6 weeks):
    text conversion: AGRKB:101, AGRKB:102
```

## Test plan

- [x] `flake8` (project config) — clean; `process_no_parent` stays under the complexity limit
- [x] `mypy --config-file mypy.config` — no issues
- [x] `report_counts` exercised in isolation with stub `counts` + `failed_refs`; section renders per MOD/phase, matches the failed counts, and is appended to `slack_messages`
- [ ] Debug dry-run on a dev DB: `python -m agr_literature_service.lit_processing.data_check.check_wft_in_progress -d True` — confirm the failed-references section matches the "Setting to failed for ..." lines
- [ ] Real run on dev; confirm the section appears in `{mod}_check_workflow_problems.log`

### DIFF
--- a/agr_literature_service/lit_processing/data_check/check_wft_in_progress.py
+++ b/agr_literature_service/lit_processing/data_check/check_wft_in_progress.py
@@ -241,17 +241,20 @@ def process_parent(db_session, phase, slack_messages, counts, failed_refs, debug
                 counts[wft.mod_id][phase['name']]['failed'] += 1
                 failed_refs[wft.mod_id][phase['name']].append(get_curie(reference, wft))
             if not debug:
-                if phase['slack message']:
-                    if wft.mod_id not in slack_messages:
-                        slack_messages[wft.mod_id] = []
-                    try:
-                        slack_messages[wft.mod_id].append(
-                            f"Setting {reference} to needed failed for {atp_get_name(wft.workflow_tag_id)}")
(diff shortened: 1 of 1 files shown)


sweng66 (approved):
Thanks @ianlongden

PR #1239, opened 2026-07-14
feat(embedding): SCRUM-6144 classifier-embedding backfill script

valearna:
## SCRUM-6144 — embed FB training set references

Adds a one-off backfill script, `agr_literature_service/lit_processing/oneoff_scripts/embed_training_set_references.py`, that generates + registers classifier embeddings (SCRUM-6142) for a MOD's references, reusing the existing per-reference machinery (no reimplementation).

### What it does
For each target reference:
- **Already converted** (has a final `converted_merged_main` Markdown): generates + registers embeddings only, via `generate_classifier_embeddings_for_reference`. The parquet is stored through `embedding_file_crud.create_or_update` → `file_upload_single`, which has no file-conversion workflow-tag gate, so the upload is allowed and the file-conversion workflow tag is left untouched.
- **Not yet converted**: runs conversion first (nXML-preferred, PDFX fallback, supplements) via `process_single_reference` exactly as pdf2md does — which generates the embeddings itself once the Markdown persists. `--skip-conversion` embeds only already-converted references.

### Flexibility
- `--mod FB [WB ...]` — one or more MODs
- `--scope training` (topic-classifier training sets, latest dataset version per topic) or `corpus` (whole MOD corpus)
- `--filter-date-before YYYY-MM-DD` — mirrors the classifier trainer's date filter (uses `date_published`, then `date_published_start`; unparseable/missing dates are kept)
- `--reference-curie CURIE ...` — target specific references
- `--no-supplements`, `--force-pdfx`, `--limit`

### Safety
- Default is a **read-only dry run** (enumerate + report); `--commit` is required to write anything.
- Fails fast on `--commit` when `OPENAI_API_KEY` is unset (otherwise embedding would silently no-op for every reference).
- Idempotent — re-runs skip already-embedded sources and already-converted mains.

Immediate use: FB training-set embeddings (SCRUM-6144). Reusable for WB (SCRUM-6143) and whole-corpus backfills.

flake8 + mypy clean.

### DIFF
--- a/agr_literature_service/lit_processing/oneoff_scripts/embed_training_set_references.py
+++ b/agr_literature_service/lit_processing/oneoff_scripts/embed_training_set_references.py
@@ -1,0 +1,446 @@
+"""SCRUM-6144: Backfill classifier embeddings for a MOD's references.
+
+Generates + registers the classifier embeddings (SCRUM-6142) for a set of
+references, reusing the exact per-reference machinery the pdf2md conversion job
+already runs, so this is a pure backfill of what conversion would have produced.
+
+Flexible by design (so the same script covers the whole epic):
+
+  --mod FB [WB ...]     one or more MOD abbreviations to process
+  --scope training      references in the MODs' topic-classifier training sets
+                        (latest dataset version per topic)  [default]
+          corpus        every reference in the MODs' corpus
+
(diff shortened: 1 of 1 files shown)


sweng66 (approved):
Thanks @valearna

PR #1241, opened 2026-07-14
refactor(laboratory): split by_name_or_strain_designation into two en…

azurebrd:
…dpoints

The combined /laboratory/by_name_or_strain_designation endpoint conflated a name substring search with a strain-code lookup behind one query param. Since strain designations are short (often two-letter) codes, folding both lookups together was confusing and error-prone.

Replace it with two purpose-built endpoints:
- GET /laboratory/by_name — case-insensitive substring match on name
- GET /laboratory/by_strain_designation — exact case-insensitive strain match

Factor the shared eager-load joins into _show_load_options() and update tests to cover the two endpoints, including that the lookups stay cleanly separated.

### DIFF
--- a/agr_literature_service/api/crud/laboratory_crud.py
+++ b/agr_literature_service/api/crud/laboratory_crud.py
@@ -261,22 +261,10 @@ def show(db: Session, curie_or_laboratory_id: str) -> LaboratoryModel:
     return obj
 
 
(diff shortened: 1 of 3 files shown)


valearna (approved):
LGTM, thanks @azurebrd !

azurebrd:
Thanks @valearna !

PR #1242, opened 2026-07-15
add first pass curation workflow tags to Workflow facet…

sweng66:
… (SCRUM-6287)

Add the "first pass curation" (ATP:0000329) process as a new Workflow facet parent so FB curators can find papers by their first-pass and community curation manual workflow tags. Descendant status tags (needed, in progress, blocked, finished, TBD) are expanded via get_atp_ids.


valearna (approved):

sweng66:
Thanks @valearna

PR #1243, opened 2026-07-17
scope confidence-level TET exclusion to the selected topic

sweng66:
Excluding a confidence level (e.g. NEG) previously attached a must_not only inside positive nested TET queries, so it did nothing when no positive TET facet was selected (the filter block was also dropped because the whole-ref guard ignored confidence_level), and it kept references that had a NEG tag for the selected topic as long as one matching tag was not NEG.

Emit the exclusion as a top-level nested must_not that combines the excluded level with the selected topic/entity terms, so a reference is dropped only when it has a tag that is both the selected topic AND the excluded level. A NEG tag on a different topic no longer removes the reference (a plain whole-reference negation dropped almost everything, since NEG tags are ubiquitous across topics). The confidence_score range is excluded from the scope, and with no positive TET facet the exclusion falls back to any tag at that level. Source method / SEA exclusions stay global whole-reference. Include confidence_level in the has_tet_whole_ref_negation guard so a standalone exclusion keeps its filter. Tests updated to the topic-scoped semantics.

### DIFF
--- a/agr_literature_service/api/crud/search_crud.py
+++ b/agr_literature_service/api/crud/search_crud.py
@@ -625,14 +625,16 @@ def search_references(
         date_created=date_created,
     )
 
-    # Remove empty filter if no facets/dates were added. A standalone source method / SEA
-    # exclusion adds a top-level must_not nested clause, so keep the filter in that case.
+    # Remove empty filter if no facets/dates were added. A standalone confidence level /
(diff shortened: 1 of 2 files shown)


valearna (approved):
Not sure how the changes work @sweng66 but I'm going to approve the PR. Maybe we can test it together at standup?

PR #1245, opened 2026-07-17
SCRUM-5781: add ABC-embedding recipe columns to ml_model

valearna:
## Summary
Supports SCRUM-5781 (retrain FB classifiers on the ABC's precomputed embeddings). Adds two nullable columns to `ml_model` so a model self-describes whether it was trained on ABC embeddings and which stored embedding to fetch:

- `embedding_profile` (String, indexed) — the ABC embedding profile; **NULL for legacy BioWordVec models**.
- `embedding_version` (Integer).

That `(profile, version)` pair is all the classifier needs *before* fetching, to select which stored embedding parquet to pull for a reference. Pooling (L2 chunk-mean) and the BoW block are fixed conventions for every ABC-embedding model (hard-coded in the consumer, not stored), and `embedding_model`/`embedding_dim` are read from the parquet's own recipe metadata — so they are intentionally not columns.

## Changes
- `ml_model` model: the two nullable columns.
- Wired through the create schema (`MLModelSchemaBase`), the `/ml_model/upload` endpoint (Form params), and the crud (persist + return in `get_model_schema_from_orm`).
- Alembic migration (hand-trimmed to only these three ops — the autogenerate had picked up unrelated drift on other tables, which was removed).

## Consumer
The consuming side lives in `agr_automated_information_extraction` (branch `SCRUM-5781`): the trainer stamps `(profile, version)` at upload; the classifier routes per model on `embedding_profile` being set, and falls back to the unchanged BioWordVec path when NULL. Backward-compatible — legacy models leave the columns NULL.

## Deploy note
Deploy this (migration + endpoint) to stage before any ABC-embedding model is promoted to `production=true`.

### DIFF
--- a/agr_literature_service/api/crud/ml_model_crud.py
+++ b/agr_literature_service/api/crud/ml_model_crud.py
@@ -79,7 +79,13 @@ def upload(db: Session, request: MLModelSchemaPost, file: UploadFile):
         data_novelty=request.data_novelty,
         negated=request.negated,
         file_classes=request.file_classes,
-        description=request.description
+        description=request.description,
+        embedding_profile=request.embedding_profile,
+        embedding_version=request.embedding_version,
+        embedding_model=request.embedding_model,
+        embedding_dim=request.embedding_dim,
+        embedding_pooling=request.embedding_pooling,
+        use_bow_features=request.use_bow_features
     )
     try:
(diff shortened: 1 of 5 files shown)


sweng66 (approved):
Thanks @valearna

PR #1251, opened 2026-07-22
Scrum 6311

azurebrd:


azurebrd:
@claude review this

azurebrd:
@claude is this right

  On finding #1 (startup warm-up "only warms one worker") — this one is worth a correction, because both reviewers analyzed setup.py in isolation and
  missed the gunicorn detail: setup_database() runs in the on_starting hook (master process) with preload_app=True, before workers fork. So the master
  wins the lock, load_initial() warms its cache, and every forked worker inherits that warm snapshot copy-on-write — workers do not start empty in
  the real deployment. So it's not actually broken as deployed. Moving load_initial() outside the lock branch would be a harmless one-line defensive
  change for any non-preload launch, but it's optional, not a fix for a live bug.

  On #2/#3 (lock held across the fetch; sync I/O on the request path) — known and accepted: bounded by the 15-min TTL + the 5s×2 timeout + 60s
  backoff, so it's a ~once-per-15-min single-thread blip, and still cheaper than the old per-request DB query. The clean improvement is
  fetch-outside-the-lock (readers keep serving stale while one refresher does I/O) or a background refresher — I deliberately deferred that as future
  work rather than add concurrency machinery this PR doesn't need.

sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #1252, opened 2026-07-23
fix(crud): harden ml_model download + fix TET NULL-PK & MCA savepoint warnings

ianlongden:
## Summary

Fixes three sources of production `ERR` log noise (and the real bugs behind them) in the
CRUD layer, surfaced by the entity-extraction pipeline and prod logs.

## 1. `ml_model/download` returned unhandled 500s (`ml_model_crud.py`)

`download_model_file` could return a raw 500 in several cases. Hardened:

- **Concurrency race (likely cause of the pipeline 500s):** it wrote the gzipped download
  and its decompressed output to **bare relative filenames** derived only from the version
  number (`<version>.gz` / `<version>.<ext>`). Parallel downloads of the same model collided —
  one request's `os.remove`/cleanup deleted the file another was still reading. Now uses a
  unique per-request `tempfile.mkdtemp()` directory.
- **S3 errors → 502:** exceptions `download_file_from_s3` doesn't swallow (`BotoCoreError`/
  `ClientError` — credentials, connectivity) now return 502 instead of 500.
- **Corrupt / non-gzip object → 502:** `gzip.BadGzipFile`/`OSError`/`EOFError` during
  decompression now return 502 instead of 500.
- Temp dir is cleaned up on every error path; on success it's removed by the response
  background task. Missing S3 object still returns **404** (unchanged).
- Added tests for the corrupt-gzip and S3-error 502 paths (local, via monkeypatch — no AWS needed).

## 2. `topic_entity_tag_crud.show_tag` NULL primary-key warning

`if 'ml_model_id' in topic_entity_tag_data:` guarded on **key presence**, but the value is
often `None` (TETs without a model), so `db.query(MLModel).get(None)` emitted SQLAlchemy's
*"fully NULL primary key identity cannot load any object"* warning. Now guards on the value
being non-None and uses the 2.0-style `db.get(MLModel, id)`.

## 3. `mod_corpus_association.batch_update_corpus` broken savepoint

The batch loop opened a per-item `db.begin_nested()` SAVEPOINT and closed it with
`nested.commit()`, but the workflow/TET/xref helpers it calls each issue their own
`db.commit()`, which ends the outer transaction and detaches the savepoint — producing
*"nested transaction already deassociated from connection"* and making the savepoint a no-op
(no real per-item isolation). Replaced with a per-item `db.commit()` / `db.rollback()`,
matching what the helpers already do, and documented why.

## Testing

- `flake8` + `mypy` clean on all changed files.
- 3 `ml_model` download tests pass on PostgreSQL 17 (existing 404 + two new 502 tests).
- **Not runtime-verified:** the `show_tag` and `batch_update_corpus` paths are Cognito/network-gated,
  so those changes (behavior-preserving) are covered by lint/type + reasoning, and will be
  exercised by CI.

### DIFF
--- a/agr_literature_service/api/crud/ml_model_crud.py
+++ b/agr_literature_service/api/crud/ml_model_crud.py
@@ -1,9 +1,12 @@
 import gzip
+import logging
 import os
 import shutil
+import tempfile
 from typing import Optional
 
 import boto3
+from botocore.exceptions import BotoCoreError, ClientError
 from fastapi import UploadFile, HTTPException
 from sqlalchemy.orm import Session, joinedload
 from starlette.background import BackgroundTask
@@ -21,6 +24,7 @@ from agr_literature_service.api.schemas.ml_model_schemas import (
 from agr_literature_service.lit_processing.utils.s3_utils import download_file_from_s3
 
 s3_client = boto3.client('s3')
+logger = logging.getLogger(__name__)
 
 
 def get_ml_model_s3_folder(task_type: str, mod_abbreviation: str, topic: str):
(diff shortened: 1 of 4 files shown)


sweng66 (approved):
Thanks @ianlongden

PR #1254, opened 2026-07-24
test(debezium): add real-time deletion propagation tests (SCRUM-6337)

valearna:
## Summary

Adds two hard-failing integration tests to the Debezium pipeline suite that verify **row deletions propagate** from PostgreSQL through Kafka/ksqlDB to Elasticsearch:

- `test_real_time_author_delete_sync` — deletes one of a reference's two authors and asserts it is removed from the reference document's `authors` array while the other remains (the joined-table `GROUP BY reference_id` aggregate delete path).
- `test_real_time_reference_delete_sync` — deletes an entire reference and asserts its document is removed from both the private and public indexes.

Each test asserts the data is present in the index first, then deletes and asserts removal, so a pass reflects genuine tombstone propagation (not data that was never indexed). Both are hard assertions (no `xfail`) and reuse the pre-populated WB MOD to avoid the `abbreviation` unique-constraint collision.

## Why

Deletion propagation was suspected to be broken during the PostgreSQL 17 / Debezium 2.7 upgrade work (SCRUM-6318). These tests exercise deletion behavior **independently of that upgrade**, on the current pipeline, so any regression the upgrade introduces is caught. See [SCRUM-6337](https://agr-jira.atlassian.net/browse/SCRUM-6337).

## Result

Both delete paths propagate correctly on the current (pre-upgrade) pipeline — which also updates SCRUM-6337: the original "deletions never propagate" conclusion does **not** reproduce here. Full `debezium-integration-test` CI run: **11 passed** ([run 30063557381](https://github.com/alliance-genome/agr_literature_service/actions/runs/30063557381)).

The tests are marked `@pytest.mark.debezium @pytest.mark.webtest`, so they run only in the Debezium integration CI job and are excluded from the default unit-test run.

[SCRUM-6337]: https://agr-jira.atlassian.net/browse/SCRUM-6337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_debezium_integration.py
+++ b/tests/test_debezium_integration.py
@@ -376,6 +376,187 @@ class TestDebeziumIntegration:
 
     @pytest.mark.debezium
     @pytest.mark.webtest
+    def test_real_time_author_delete_sync(self, db, mock_data_factory, elasticsearch_config):  # noqa
+        """Deleting a joined-table row must be reflected in the reference's ES document.
+
+        Creates a reference with two authors, waits for both to appear in the private
+        index, then deletes one author and asserts it is removed from the reference's
+        ``authors`` array while the other author remains.
+
+        This exercises the joined-table delete path (``author`` -> ``GROUP BY
+        reference_id`` aggregate). If a row deletion is not propagated, the deleted
+        author lingers in the document and this test fails. See SCRUM-6337.
+        """
+        import time
+
+        test_curie = "AGRKB:101000998"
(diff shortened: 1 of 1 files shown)


azurebrd (approved):
Thanks Valerio

PR #1257, opened 2026-07-27
propagate resolved evidence-assertion name to grid entries

sweng66:
Include source_evidence_assertion_name (resolved from the ateam ontology store) on each aggregated grid mini-row so the UI can render the real label instead of relying on a hardcoded curie->label list.


valearna (approved):

sweng66:
Thanks @valearna

PR #1258, opened 2026-07-30
fix(pg17): correct terminal handling in containerized client wrappers (SCRUM-6374)

valearna:
## Context

[SCRUM-6374](https://agr-jira.atlassian.net/browse/SCRUM-6374)

The dev EC2 box is Amazon Linux 2 (EL7) and ships a PG13 client, which cannot dump the now-PG17 RDS servers. The PG17 client tools were therefore wrapped to run out of the `postgres:17-alpine` image, with symlinks in `/usr/local/bin`. A container only has a controlling terminal when docker allocates a pty, and that broke interactive password prompts in two different ways.

## Problems

**`pg_dump ... > file`** wrote the password prompt *into the output file* and appeared to hang. The wrapper allocated a pty whenever **stdin** was a terminal — but `pg_dump > file` redirects only stdout, so stdin was still the terminal. A pty collapses stdout, stderr and `/dev/tty` into one stream, which the shell redirect then captured. pty newline translation (LF → CRLF) would also have silently corrupted binary `-Fc` dumps.

**`pg_restore ... < dump`** failed with `password authentication failed`. With no controlling terminal, libpq falls back to reading the password from stdin — so it consumed the archive's first line (the `PGDMP` header) and sent that as the password. It would have failed even with the right password, since the archive was then short a line.

## Fix

Terminal handling is split by which fds are actually available:

| stdin | stdout | mechanism |
|---|---|---|
| tty | tty | allocate a pty (interactive `psql`) |
| data | tty | bind-mount the real terminal onto `/dev/tty`, so libpq prompts there and leaves stdin intact |
| tty | redirected | do nothing — libpq's stdin fallback works here |

That last row matters: mounting `/dev/tty` in that case *actively breaks* it, because the docker client is itself reading the same terminal to forward stdin and consumes the keystrokes before the container can.

## Verification

Driven against a real pty (`pty.fork`) rather than an ssh harness, which confounds the very fds under test:

- `pg_dump -W > file` — prompt on terminal, password delivered, **0 bytes** leaked into the file on auth failure
- `pg_restore -W < dump` — prompt with echo suppressed, custom-format archive read from stdin, data landed (`mod rows=9`, `users rows=1527`)
- interactive `psql` — real pty, prompt and echo suppression normal
- redirected `pg_dump` output is **byte-identical** to `-f` output (435462 bytes both, differing only in the embedded timestamp)
- `pg_dump -Fd -j 4` — parallel directory dump created and read back via `pg_restore --list`

Known caveat, inherent rather than fixable: in the stdin-is-a-tty / stdout-redirected case the password is **visible as you type**, because echo cannot be disabled on a stdin that docker is forwarding. Prefer `PGPASSWORD` or `~/.pgpass` for anything scripted.

## Also: parallelized `sync_data.sh`

Dump and restore now use multiple jobs by default, scaled to the host and capped at 8. Parallel dumps use directory format because `pg_dump -j` requires `-Fd`; `--jobs N` tunes it and `--jobs 1` restores the previous serial single-file behaviour. Cleanup and size reporting handle both shapes.

Note `sync_data.sh` was never affected by the prompt bugs — it sets `PGPASSWORD` and writes with `-f` rather than a shell redirect, so it never prompts and never writes through stdout.

[SCRUM-6374]: https://agr-jira.atlassian.net/browse/SCRUM-6374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/install_pg17_wrappers.sh
+++ b/install_pg17_wrappers.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+# Installs PG17 client wrappers backed by the postgres:17 docker image.
+#
+# Why: this box is Amazon Linux 2 (EL7). There is no native PG17 client for it
+# (no amazon-linux-extras postgres topic; PGDG has no rhel-7 build for 17), and
+# the RDS servers are now PG17 -- the host's PG13 pg_dump can no longer dump a
+# v17 server. These wrappers put pg_dump/pg_restore/psql/... in /usr/local/bin
+# (which precedes /usr/bin in PATH), each dispatching to postgres:17-alpine.
+# This fixes sync_data.sh and every other caller with no script edits.
+#
+# Run:      sudo bash ~/install_pg17_wrappers.sh
+# Undo:     sudo rm /usr/local/bin/pg-docker-wrapper \
+#              /usr/local/bin/{pg_dump,pg_restore,psql,pg_dumpall,pg_isready,\
+#              vacuumdb,reindexdb,createdb,dropdb,createuser,dropuser,clusterdb}
+set -eo pipefail
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo "ERROR: run this with sudo:  sudo bash $0" >&2
+    exit 1
+fi
+
+IMAGE="postgres:17-alpine"
+WRAPPER=/usr/local/bin/pg-docker-wrapper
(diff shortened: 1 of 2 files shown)


sweng66 (approved):
Thanks @valearna

valearna:
Thanks @sweng66 !

PR #1260, opened 2026-08-01
bound per-reference TET tags in search index to avoid ES nested-limit halt

sweng66:
topic_entity_tags is an Elasticsearch `nested` field, capped at 10000 sub-documents per doc by index.mapping.nested_objects.limit (default). A handful of bulk ZFIN papers carry 10k-50k gene/allele TET tags, so ksqlDB's uncapped COLLECT_LIST emitted >10k-element arrays; ES rejected those documents and the single-partition, fail-fast ES sink task died, silently halting the entire reindex. This left gene tags partially indexed and allele tags (loaded last) entirely absent from search.

- docker-compose.yaml: set KSQL_KSQL_FUNCTIONS_COLLECT_LIST_LIMIT=9000 so every per-reference COLLECT_LIST stays under the ES nested limit. The database keeps all associations; only the search document is bounded.
- elasticsearch-sink.json: add errors.tolerance=all, warn on malformed documents, and a dead-letter queue as a backstop so a stray oversized record is diverted and logged instead of halting the whole reindex.


sweng66:
can you check again? @claude

sweng66:
Check again? @claude

sweng66:
@claude check again?

PR #1262, opened 2026-08-03
Scrum 6311

azurebrd:


azurebrd:
@claude review

azurebrd:
@claude  review again.  resource descriptors for SGD and ORCID are a future issue.  we don't have a data issue on prod db for 3   2 the sript is there for historical purposes, it won't be run again and there's notes  4 is fine, downgrade could break,but it shouldn't be needed  5  measured the live DB: max 17 members/lab, 21 lineage rows/person, against a 25-row term table, all PK lookups. Negligible, so left
  as-is with a scale note on serialize_term_ref documenting the numbers and the selectinload/batch fix to reach for if lists ever grow.

azurebrd:
@claude review

azurebrd:
@claude review again

azurebrd:
@claude review again.    Finding 1 (enum slugs not in the mapping) — no change, by design. We confirmed only 3 of the 14 LabPosition slugs (postdoc, graduate_student,
  research_professor) are in the curator's flatfile; the other 11 were never in the curator's source, and zero of them exist in the data on stage or
  prod. Since the flatfile is the curator's authoritative source and the DB doesn't contain those values, we're not hand-authoring mappings for them.
  If one is entered via the current UI before deploy, the fail-closed guard aborts with the exact value named and the deploy re-runs after the curator
  adds it — a stopped deploy, never corrupted data. (This PR also deletes the LabPosition enum, so the window closes at deploy.) If the curator wants
  the window shut proactively, they add the 11 values to the flatfile and we re-pull.

  Finding 2 (CI never runs the migration) — acknowledged, intentionally skipped. Correct that the missing_terms guard first executes at deploy. But
  this revision runs exactly once per DB, the invariant is green today, and our databases are created by restoring dumps (4002 came from a prod dump;
  the test DB is built from models) rather than by replaying the migration chain — so a future vocabulary_seed_data.py rename couldn't re-trigger this
  migration on any existing environment. The only exposure is a fresh-from-migration-replay build after such a rename, which isn't part of our
  workflow. Not worth importing a migration into the test suite to guard.

  Finding 3 (is_pi stamped despite former_pi/alum) — fixed (e0a08924). Real but tiny: 0 rows on prod, 1 on stage (Eric Phizicky, id 4416 — since
  removed; it was a manual test edit). Rather than silently skip, we added a fail-closed abort: if any PI-bucket row is former_pi/alum with is_pi
  NULL, the migration raises and a curator resolves the contradiction. Verified 0 such rows on stage and prod, so it passes cleanly on the real
  targets and only fires on genuinely contradictory data.

  Finding 4 (free text unrecoverable in-DB) — no change, deliberate. We're keeping the no-archive decision. The hardcoded mapping in the migration is
  the auditable record of the transform, and the odd-looking collapses are the curator's judgment on their own corpus. We accept that re-deriving
  would require a backup restore.

  Finding 5 (stale "archive" wording) — fixed (e0a08924). Docstring title now says "lab_position mapping" (with a note that the filename encodes the
  immutable revision id), and the step-4 tombstone no longer references a non-existent archive.

  On the DELETE removal / laboratory_person 5620: confirmed still present on live prod (Ceri Van Slyke → Elizabeth Grayhack Lab, created 2026-07-29,
  after the original survey) — thanks for pushing on that.

  Validation: ran the full migration on a prod-restored 4002 — 642 rows mapped to a role FK with a term-for-term match to the pre-run dry-run, 1130
  PI-bucket rows flagged is_pi, nothing deleted, no guard tripped. The finding-3 guard's exact expanding-bindparam path was exercised read-only
  against stage (0 conflicts). We do not plan to reload prod/stage into 4002 to re-test these two follow-up edits, since finding 5 is comment-only and
  finding 3 is a read-only guard already confirmed to return 0 on both targets.

azurebrd:
Thanks @sweng66 !

